PR #81, opened 2021-08-07
AkanshaJain_Assignment_6

akansha-jain6:
Assignment 6 Screenshot 2021-08-07 at 8 10 24 PM


harshvegada (approved):
If you want to use a variable at multiple places then you should instance variable, #best practices

PR #87, opened 2021-08-07
Sameer_Assignment6

SZalke:
image


harshvegada (approved):
If you want to use a variable at multiple places then you should instance variable,

PR #98, opened 2021-08-07
Bhakti Kakade Assignment7

bhaktibk:
Simple interest is : 860708.9
====================================
Temperature is : 112.47222 Celsius degree


harshvegada (approved):
Code Looks Good to me

PR #119, opened 2021-08-07
Nasir_Assignment6

nasiruddin16:
image
image


harshvegada (approved):
If you want to use a variable at multiple places then you should instance variable, 
method and variable name should be in small letters 
method name should be meaningful 
#best_practices

nasiruddin16:
Changes are done as per the comments

PR #121, opened 2021-08-07
Akshay_Assignment_6_Updated

akshay0795:
image


harshvegada (approved):

akshay0795:
> method and variable name should be in small letters #best_practices

I will follow the same from now onwards, Thanks!!

PR #145, opened 2021-08-08
Nidhi_Assignment_6

Nidy-C:
image

### DIFF
--- a/src/nidhi/Assignment_6.java
+++ b/src/nidhi/Assignment_6.java
@@ -1,0 +1,20 @@
+package nidhi;
+
+public class Assignment_6 {
+		int simpleInt;
+		int cel;
+	void simpleInterest(int p, int t, int r) {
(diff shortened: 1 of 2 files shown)


src/nidhi/Assignment_6.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class Assignment_6 {
+		int simpleInt;
+		int cel;
+	void simpleInterest(int p, int t, int r) {
+		simpleInt = (p * r * t)/100;
+		System.out.println(simpleInt);
+	}
+	
+	void fahToCel(int f) {
```
KrishnaKTechnocredits, line 11: Write full method name.

Nidy-C: ok, will do it

KrishnaKTechnocredits: Done.

PR #153, opened 2021-08-08
Purshottam_Assignment6

purshottam94:
image


harshvegada (approved):
method and variable name should be in small letters,

PR #177, opened 2021-08-08
Anuja_Assignment_7

AnujaTechno:
image
image
image
image
image


harshvegada (approved):
Logic looks good

points to improve 
1. space should not be more in methods or inside method
2. code indentation should be maintain


Kindly take looks of below Link

https://github.com/KrishnaKTechnocredits/JAVATechnoJuly2021/pull/174/files

PR #235, opened 2021-08-09
Priti_Assingment_7

psnilange:
Assingment_7 -for loop 
image


KrishnaKTechnocredits (approved):
Logic looks good to me!
method name should be in small letters #BestPractice

PR #243, opened 2021-08-09
Chandani assignment 7

chandni0604:
image
image

### DIFF
--- a/src/chandani/Assignment7.java
+++ b/src/chandani/Assignment7.java
@@ -1,0 +1,88 @@
+package chandani;
+
+class Assignment7{
+	
+	int index = 0;
+	void findEvenOdd(int startindex, int endindex){
+		System.out.println("Even numbers are ");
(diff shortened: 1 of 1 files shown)


src/chandani/Assignment7.java
```diff
@@ -0,0 +1,88 @@
+package chandani;
+
+class Assignment7{
+	
+	int index = 0;
```
KrishnaKTechnocredits, line 5: this is common, it should not reason is you calling a method which increases index value and again on same class reference you are calling second method which also use same index variable 
this cause issue. Think on this. 

you should create local variable it's oky.

KrishnaKTechnocredits (approved):
Logic looks good
take of comment next time

src/chandani/Assignment7.java
chandni0604: I'll take care of this.

PR #246, opened 2021-08-09
ChaitanyaD_Assignment7

chaitu2794:
image
image

### DIFF
--- a/src/chaitanya/ass7ManyMethods.java
+++ b/src/chaitanya/ass7ManyMethods.java
@@ -1,0 +1,154 @@
+//Assignment - 7 : 8th Aug2021
+
+//Write only one program having following methods. [7 methods in same program] 
+
+//1. On user define range print all even numbers. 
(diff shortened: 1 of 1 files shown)


src/chaitanya/ass7ManyMethods.java
```diff
@@ -0,0 +1,154 @@
...
+//         15
+//         13
+//         11
+
+
+package chaitanya;
+
+public class ass7ManyMethods {
+
+	int index;
```
KrishnaKTechnocredits, line 64: this variable should not be at instance, reason is think on same reference variable you are calling one method update index value and on same reference variable you call second method not that method have index value which last increase by first method.

KrishnaKTechnocredits: Think on this else we can connect on skype this.
note PR number while discussing.

KrishnaKTechnocredits (approved):
Code looks good to me, 
Keep it up!

PR #262, opened 2021-08-10
Precilla_Assignment8

precillaaranha16:
image


src/precilla/Assignment8.java
```diff
@@ -0,0 +1,64 @@
+package precilla;
+
+public class Assignment8 {
+	float currentBalance;
+	String userName;
+	int debitCount, creditCount, printBal;
+	static int debitCnt, creditCnt ,printBalance;
```
KrishnaKTechnocredits, line 7: you should give totalDebitCnt or something so it will clear to all for understanding adn not get confused between above variable.

KrishnaKTechnocredits (approved):
Code looks good to me,  
Keep it up!

PR #266, opened 2021-08-10
Shraddha_Assignment_8

shraddhaatal:
image

### DIFF
--- a/src/shraddha/BankingTransaction.java
+++ b/src/shraddha/BankingTransaction.java
@@ -1,0 +1,101 @@
+package shraddha;
+
+public class BankingTransaction {
+	int creditAmt;
+	int debitAmt;
+	int balance;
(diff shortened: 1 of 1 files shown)


src/shraddha/BankingTransaction.java
```diff
@@ -0,0 +1,101 @@
+package shraddha;
+
+public class BankingTransaction {
+	int creditAmt;
+	int debitAmt;
+	int balance;
+	int debitCount;
+	int creditCount;
+	int balanceCount;
+	static int debitCounter;
```
KrishnaKTechnocredits, line 10: you should give totalDebitCnt or something so it will clear to all for understanding adn not get confused between above variable.

KrishnaKTechnocredits (approved):
Code looks good to me, 
Keep it up!

PR #302, opened 2021-08-11
Bhararat_assignment8

bharattanpure:
assignment 8


a-malvi (approved):
code looks good, 
Suggestion : avoid spelling mistakes like "total" is "totle" rest all good.

PR #307, opened 2021-08-11
Mratunjay_Assignment_8

Mratunjay24:
image

### DIFF
--- a/src/mratunjay/Assignment_8.java
+++ b/src/mratunjay/Assignment_8.java
@@ -1,0 +1,75 @@
+package mratunjay;
+
+public class Assignment_8 {
+
+	int PrintCount = 0;
+	int DebitCount = 0;
+	int CreditCount = 0;
+
+	static int PrintAllCount = 0;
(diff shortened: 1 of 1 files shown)


src/mratunjay/Assignment_8.java
```diff
@@ -0,0 +1,75 @@
+package mratunjay;
+
+public class Assignment_8 {
+
+	int PrintCount = 0;
```
harshvegada, line 5: Variable name should starts with small letters

SandeshMolane (changes requested):
Please follow proper indentation. Before submitting code for review you can use Ctrl + Shift + F shortcut keys for indentation. As per Java best practice First alphabet of Method name should be small case. The code looks logically correct but one problem sub part is missed. It is to print remaining balance once user withdraws some amount. you can refer - https://github.com/KrishnaKTechnocredits/JAVATechnoJuly2021/pull/376/files

harshvegada (approved):
work on comments,
Rest code looks good!

PR #318, opened 2021-08-11
Santosh_Assignment_8

Santoshghige:
image


nehatawari:
image
it should be creditCount2++ and not creditcount2+2.for every transaction it will increase by 1.

Santoshghige:
ok mam.

nehatawari:
change code and add commit and push again

PR #320, opened 2021-08-11
Umakant_Assignment_9

Umakant0402:
image

### DIFF
--- a/src/umakant/Shop.java
+++ b/src/umakant/Shop.java
@@ -44,47 +44,50 @@ public class Shop {
 		if (maggieCount > 0 || dosaCount > 0 || oilCount > 0 || panipuriCount > 0 || masalaCount > 0) {
 			System.out.println("Below items are available:");
 			if (maggieCount > 0)
-				System.out.println("Maggie -"+maggieCount +" Packets");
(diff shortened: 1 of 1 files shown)


src/umakant/Shop.java
```diff
@@ -0,0 +1,106 @@
...
+	static void setInitialQuantity(int maggie, int dosa, int oil, int panipuri, int masala) {
+		maggieCount = maggie;
+		dosaCount = dosa;
+		oilCount = oil;
+		panipuriCount = panipuri;
+		masalaCount = masala;
+	}
+
+	static void outOfStockItems() {
+		if (maggieCount == 0 || dosaCount == 0 || oilCount == 0 || panipuriCount == 0 || masalaCount == 0) {
```
harshvegada, line 35: this condition not require! first if condition

harshvegada (approved):
Code looks good to me!
Keep it up @Umakant0402

src/umakant/Shop.java
Umakant0402: Without this if how we can print else message i.e. how we can check no item are out of stock.

PR #323, opened 2021-08-11
Elizabeth_Assignment_9

Elizmary:
image


a-malvi:
@Elizmary : Expectation is to create two separate methods, it seems you mixed it into single. Please correct code as per requirement.

Elizmary:
> @Elizmary : Expectation is to create two separate methods, it seems you mixed it into single. Please correct code as per requirement.

@a-malvi Have divided that method into 2 separate methods. Please review. Also attached the output after correction

a-malvi:
Code looks good, however rather than hardcoding of initial quantity you can make then define by user and also mark them static so there values should remain intact whenever utilized by any method.
Good to merge.

PR #324, opened 2021-08-11
AmolVyas_Assignment_9

amolvyas2311:
image


VaibhavTati:
In this case without static variable also it will work.
Overall OK to merge.

PR #328, opened 2021-08-11
Neha_Assignment_9

NehaJTechnocredits:
**Outputs with both coding approaches are same.**
_Output with code with method with return type:_
image
image

_Output for alternate code:_
image
image


harshvegada (approved):
both approaches looks good, first would be choice for this.
Code looks good to me!

PR #355, opened 2021-08-11
Sagar_Assignment9

sagaanna:
image


nehatawari:
use modular approach means method for out of stock method for available stock.right now which stock is running out of stock is not clear.

PR #356, opened 2021-08-11
JagratiB_Assignment_9

Jagrati31:
image


Jagrati31:
Thank you Sangeeta for review comments ... Will definitely work on that  .
 but Can you review my AvailableInStock and outOf Stock method one more time . it is not working as per expected out put .
Can you please let me know what changes required on that  ?

PR #357, opened 2021-08-11
Precilla_Assignment9

precillaaranha16:
image

### DIFF
--- a/src/precilla/Shop.java
+++ b/src/precilla/Shop.java
@@ -1,0 +1,107 @@
+package precilla;
+
+public class Shop {
+	static int maggiePackets,dosaPackets,pouchesPackets,panipuriPackets,masalasPackets;
+	String choice;
+	void initializeStocksOfPackets() {
+		maggiePackets=0;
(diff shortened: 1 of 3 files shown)


src/precilla/staticExample.java
```diff
@@ -0,0 +1,15 @@
+package precilla;
+
+public class staticExample {
```
SandeshMolane, line 3: class name should begin with Capital letter.

src/precilla/Shop.java
```diff
@@ -0,0 +1,107 @@
...
+				System.out.println(choice);
+			}else {
+				System.out.println("We are Running out of "+choice+ "!! Please shop some other items");	
+			}	
+		}else {
+			System.out.println("Given Item is not available in this shop");
+		}
+	}
+	void outofStockItems() {
+		 	if(maggiePackets <=0)
```
SandeshMolane, line 64: maggiePackets or any other item can never become less than zero so maggiePackets == 0 will also work.

PR #359, opened 2021-08-11
ShwetaD_Assignment_8

ShwetaDharmadhikari:
image


src/shweta_Dharmadhikari/static_example/Assignment8.java
```diff
@@ -0,0 +1,73 @@
+package shweta_Dharmadhikari.static_example;
+
+public class Assignment8 {
+
+	int totalbalance;
```
Upasana1703, line 5: naming convention should be proper (totalBalance) it should start with small letters following caps

Upasana1703 (approved):
code looks good to me.

PR #360, opened 2021-08-11
Nasir_Assignment9

nasiruddin16:
image

### DIFF
--- a/src/nasir/ShopSystem.java
+++ b/src/nasir/ShopSystem.java
@@ -1,0 +1,97 @@
+/*Create a system (Shop) which maintains each item's quantity.
+
+a. Initial stocks of Maggie: 50 Maggie packets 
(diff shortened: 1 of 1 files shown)


src/nasir/ShopSystem.java
```diff
@@ -0,0 +1,97 @@
...
+			System.out.println("out of stock");
+			System.out.println("running out of stock");
+		} 
+	}
+
+	public static void main(String[] args) {
+		ShopSystem shopSystem = new ShopSystem();
+		shopSystem.Stocks(80, 90, 100, 50, 70);
+		shopSystem.availableStocks(70, 100, 45, 5, 4);
+
```
harshvegada, line 94: you should write method which tells u available stock

harshvegada (approved):
Looks good to me!

PR #361, opened 2021-08-11
Sameer_Assignment_9

SZalke:
image


nehatawari:
Try to cover negative scenarios e.g if customer need 40 panipuri and shopkeeper have 30.then customer want full quantity or nothing

PR #364, opened 2021-08-11
Bhakti k assignment 9

bhaktibk:
image

### DIFF
--- a/src/bhakti/AssignmentBk_6.java
+++ b/src/bhakti/AssignmentBk_6.java
@@ -1,6 +1,6 @@
 package bhakti;
 
-public class Assignment7 {
+public class AssignmentBk_6 {
 	
 	void simpleInt(float p, int t, float r){
 		float simInt=(p * t * r)/100;
@@ -14,9 +14,9 @@ public class Assignment7 {
 	}
 	
 	public static void main(String [] rts) {
-		Assignment7 assignment7= new Assignment7();
-		assignment7.simpleInt( 34546.56f, 2019, 1.234f);
-		assignment7.celsius(234.45f);
+		AssignmentBk_6 assignmentBk_6= new AssignmentBk_6();
+		assignmentBk_6.simpleInt( 34546.56f, 2019, 1.234f);
+		assignmentBk_6.celsius(234.45f);
 		
 	}
 


src/bhakti/AssignmentBk_9.java
```diff
@@ -0,0 +1,112 @@
+package bhakti;
+
```
MadhaviKhasbage, line 2: Add assignment text here, so it will help you after 2-3 month also

src/bhakti/AssignmentBk_9.java
```diff
@@ -0,0 +1,112 @@
...
+class AssignmentBk_9{
+	int maggieQnt= 50;
+	int dosaQnt = 43;
+	int pouchesQnt= 39;
+	int panipuriQnt= 43;
+	int masalaQnt= 73;
+	 
+	
+	void checkQnt(String product, int qnt){
```
MadhaviKhasbage, line 11: Try to give full or meaningful name like, "checkProductQuantity", so it will help to understand easily when we have number of functions.

src/bhakti/AssignmentBk_9.java
```diff
@@ -0,0 +1,112 @@
...
+			System.out.println("Masala: " +masalaQnt);
+		
+		System.out.println( "---------------------");
+	}
+	
+		
+	public static void main (String[] df){
+		AssignmentBk_9 assignmentBk_9 = new AssignmentBk_9();
+		 	
+		assignmentBk_9.checkQnt("Maggi", 77);
```
MadhaviKhasbage, line 104: try to create separate functions for all products... now when you call "chkQty" it is checking each and every if condition, so it is time consuming

MadhaviKhasbage (approved):
need some changes, rest code looks good to me!

src/bhakti/Assignment_8.java
```diff
@@ -0,0 +1,63 @@
+package bhakti;
+
```
MadhaviKhasbage, line 2: Add assignment text here

MadhaviKhasbage (commented):
Code looks good to me!

src/bhakti/AssignmentBk_9.java
bhaktibk: Thank you for your input  i will take care hence forth

src/bhakti/AssignmentBk_9.java
bhaktibk: Yes i will take care

src/bhakti/AssignmentBk_9.java
bhaktibk: Thank you for your input can u please elaborate more 
My understanding is  :
in main method user can ask for any product (Maggi or any other then we need to check if that product is available, if available then asked quality is there or not i have mentioned single method where "product" and "quantity" are provided then we are checking  the same

MadhaviKhasbage: For example if you are checking quantity for "Pouches", so it will check each and every if condition.. hope it is clear now or ping me  on skype  madhavikhasbage@hotmail.com .. we can discuss on call

if(product.equals("Dosa")){
}else if (product.equals("Panipuri")){
}
}else if (product.equals("Pouches")){
}

PR #365, opened 2021-08-11
Varun_Assignment_9

varunj822:
image

### DIFF
--- a/src/varun/Assignment_9.java
+++ b/src/varun/Assignment_9.java
@@ -8,68 +8,90 @@ public class Assignment_9 {
 	int masalaStock = 73;
 
 	void stockOfItemsAvlbl(String itemName, int itemQuantity) {
-		if(itemName.equalsIgnoreCase("Maggie")) {
-			maggieStock = maggieStock-itemQuantity;
-			if(maggieStock<0) {
+		if (itemName.equalsIgnoreCase("Maggie")) {
+			if (maggieStock < itemQuantity)
 				outOfStock(itemName);
+			else if (maggieStock >= itemQuantity) {
+				System.out.println(itemName + " is availabe with " + maggieStock + " quantity");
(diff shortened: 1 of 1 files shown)


src/varun/Assignment_9.java
```diff
@@ -0,0 +1,75 @@
...
+public class Assignment_9 {
+	int maggieStock = 50;
+	int dosaStock = 43;
+	int pouchesStock = 39;
+	int panipuriStocks = 43;
+	int masalaStock = 73;
+
+	void stockOfItemsAvlbl(String itemName, int itemQuantity) {
+		if(itemName.equalsIgnoreCase("Maggie")) {
+			maggieStock = maggieStock-itemQuantity;
```
harshvegada, line 12: Here first you should check whether we have enough stock that fulfilled the current requirement of customer?

src/varun/Assignment_9.java
```diff
@@ -0,0 +1,75 @@
...
+			if(maggieStock<0) {
+				outOfStock(itemName);
+			}
+			else if(maggieStock<=10)
+			System.out.println(itemName+ " is running out of stock with only " +maggieStock+ " quantity remaining");
```
harshvegada, line 17: why are saying if stock is less than 10 then stock is not available ?

harshvegada (changes requested):
Check requirement again and update this

Reference 
https://github.com/KrishnaKTechnocredits/JAVATechnoJuly2021/pull/377/files

src/varun/Assignment_9.java
harshvegada: Yes, similar to OOS
OK

varunj822:
image

harshvegada (approved):
Code looks good to me,
 Good to merge!

PR #368, opened 2021-08-11
Shantanu_Assignment_9

ShantanuGTechnocredits:
image


nehatawari:
Out of Stock is printed in three methods,only one method is sufficient to print that.
if someone required partial quantity e.g shopkeeper having 20 item and customer need 22.customer is agree to take 20 items that scenario is not handled.rest code looks good.keep it up

PR #372, opened 2021-08-11
Purshottam_Assignment9

purshottam94:
image


surabhi-9 (changes requested):
Rewrite the code again with proper validation like :
1. if packets are less than 0, show insufficient error
2. if requested packets are less than available packets then deduct requested quantity from finalQuantity else show message that we can give you packets (packets-totalQuantity )
and assign total packets to 0.

purshottam94:
> Rewrite the code again with proper validation like :
> 
> 1. if packets are less than 0, show insufficient error
> 2. if requested packets are less than available packets then deduct requested quantity from finalQuantity else show message that we can give you packets (packets-totalQuantity )
>    and assign total packets to 0.

I have given validation inside displayOutofStockProduct and called this method inside purchaseProcess

src/purshottamJoshi/Shop.java
```diff
@@ -0,0 +1,73 @@
...
+	static int totalDosaStock = 43;
+	static int totalOilPouches = 39;
+	static int totalPanipuriPackets = 43;
+	static int totalMasalaPackets = 73;
+	void setCustomerDetails(String customer1) {
+		customerName = customer1;
+	}
+	void purchaseProcess(int maggie,int dosa,int oil,int panipuri,int masala) {
+		displayOutofStockProduct();
+		if(maggie >  totalMaggieStock) {
```
surabhi-9, line 15: available packets will be when totalMaggieStock > maggie . so this condition is wrong

purshottam94 (commented):
updated Purshottam_Assignment9 pr 
image
@surabhi-9 @harshvegada @KrishnaKTechnocredits

src/purshottamJoshi/Shop.java
purshottam94: > available packets will be when totalMaggieStock > maggie . so this condition is wrong

Updated same.

surabhi-9 (approved):
code looks good to me

KrishnaKTechnocredits (approved):
Code looks good to me!
Keep doing well!

PR #377, opened 2021-08-11
Vivek_Assignment_9

vivekvitthalpatil:
image

### DIFF
--- a/src/vivek/Assignment9.java
+++ b/src/vivek/Assignment9.java
@@ -1,0 +1,87 @@
+package vivek;
+
+public class Assignment9 {
+
+	int maggiecnt;
+	int dosacnt;
+	int oilcnt;
+	int panipuricnt;
+	int masalacnt;
+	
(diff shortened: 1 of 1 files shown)


src/vivek/Assignment9.java
```diff
@@ -0,0 +1,87 @@
...
+		panipuricnt=panipuri;
+		masalacnt=masala;
+	}
+	
+	void purchaseItems(int maggie,int dosa, int oil, int panipuri, int masala) {
+		if (maggie<=maggiecnt)
+			maggiecnt=maggiecnt-maggie;
+		else {
+			System.out.println("Maggie running out of stock");
+			maggiecnt=0;
```
harshvegada, line 24: Making maggiecnt = 0, is not require 
think If I requested u to give me 51 and u have only 50 then you are not going to make your available stock as 0 right?

harshvegada (approved):
Looks good, think on comments

PR #379, opened 2021-08-11
Priti _Assingment _9

psnilange:
image


src/priti/Assingment_9.java
```diff
@@ -0,0 +1,102 @@
...
+		}
+		if (panipuri==0) {
+			System.out.println("panipuri is out of Stock");
+		}
+	}
+		
+	void avaliableStock() {
+		
+		if (maggie != 0) {
+			System.out.println("Maggie Stock is - "+ maggie);
```
surabhi-9, line 75: condition is wrong. you should write if maggie >0 SOP();

src/priti/Assingment_9.java
```diff
@@ -0,0 +1,102 @@
...
+		if (pouchQty>pouch) {
+			System.out.println("Pouch is running out of stock");
+		}else {
+			pouch= pouch-pouchQty;
+		}
+	}
+	
+	void purchase4(int paniQty) {
+		if (paniQty>panipuri) {
+			System.out.println("Panipuri is running out of stock");
```
surabhi-9, line 47: you can also mention like we can give quantity :- paniQty-painpuri then assign 0 to panipuri. Not mandatory to apply new condition but this is extra validation we can apply

src/priti/Assingment_8.java
```diff
@@ -0,0 +1,70 @@
...
+	String customerName;
+	static int overAllDebitCnt;
+	static int overAllcreditCnt;
+		
+	void setCutomerDetails(String name){
+		customerName = name;
+	}
+	
+	void debitAmt(int amt){
+		currentBalance = currentBalance-amt;
```
surabhi-9, line 15: validation is missing while debiting amunt, e.g if(amt<=currentBalance) then only debit it else account balance will go into minus

surabhi-9 (changes requested):
Please use validation for debitting amount. rest code looks good to me

PR #381, opened 2021-08-11
Avinash_Assignment_9

AvinashSPatil:
image


src/avinash/Shop.java
```diff
@@ -0,0 +1,107 @@
...
+	void purchaseMasals(int quantity) {
+		if(quantity > initialStockofMasals) {
+			System.out.println("The required quantity of masals item is not available");
+		}else {
+			initialStockofMasals -= quantity;
+		}
+		
+	}
+	
+	void displayOutOfStockItems() {
```
MadhaviKhasbage, line 64: Here, if (intialStockofMaggie == 0) then you should say OOS, no need to check "<" for all items "=="  is correct

MadhaviKhasbage (approved):

MadhaviKhasbage:
Need small change, rest code logic looks good to me!

MadhaviKhasbage (commented):
Reviewed.

PR #384, opened 2021-08-11
Shraddha_Assignment_9

shraddhaatal:
image


nehatawari:
image
 No need to have else part which is printing same as we are printing out of stock in outOfStock method.Also if anyone need partial quantity that scenario missing. e.g if customer need 30 panipuri n shopkeeper have 25 then customer can take 25 from that shop so can add boolean flag as isPartialQuantityAccepted and on that flag codition will base...Rest Code looks Good to me.Keep it up

PR #385, opened 2021-08-11
Suruchi_Assignment_9

suruGIT21:
image

### DIFF
--- a/src/suruchi/Shop_Assignment9.java
+++ b/src/suruchi/Shop_Assignment9.java
@@ -1,0 +1,81 @@
+package suruchi;
+
+public class Shop_Assignment9 {
+
+	static int currentMaggieStock = 50;
+	static int currentDosaStock = 43;
+	static int currentOilPouchesStock = 39;
+	static int currentPanipuriStock = 43;
+	static int currentMasalaStock = 73;
+
+	void purchaseMaggie(int maggiStocks) {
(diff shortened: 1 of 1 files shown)


src/suruchi/Shop_Assignment9.java
```diff
@@ -0,0 +1,81 @@
...
+		if (currentPanipuriStock == 0)
+			System.out.println("Panipuri");
+		if (currentMasalaStock == 0)
+			System.out.println("Masala");
+	}
+
+	void displayInStockItems() {
+		System.out.println();
+		System.out.println("Below items are in stock: ");
+		if (currentMaggieStock > 3)
```
harshvegada, line 48: we don't have such requirement where we have to write such condition

suruGIT21: Okay. So I believe if we had a requirement where stock is there but shop is on the verge of running out of it, that time probably we can have this statement, otherwise we can only write the condition where if that stock is more than 0 then always print the quantity. Am I understanding this correctly?

KrishnaKTechnocredits: This is correct!

KrishnaKTechnocredits (approved):

src/suruchi/Shop_Assignment9.java
suruGIT21: Thanks Krishna!

PR #390, opened 2021-08-11
Monika_Assignment_9

MonikaZankar:
image

### DIFF
--- a/src/monika/Assignment_9.java
+++ b/src/monika/Assignment_9.java
@@ -43,7 +43,7 @@ public class Assignment_9 {
 		if(oilCount<=0)
 			System.out.println("OilPacket is running Out of Stock");
 		if(panipuriCount<=0)
-			System.out.println("Panipuri is running Out of Stock");
+			System.out.println("Panipuri Packet is running Out of Stock");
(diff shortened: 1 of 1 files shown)


src/monika/Assignment_9.java
```diff
@@ -0,0 +1,85 @@
...
+	static int maggieCount,dosaCount,oilCount ,panipuriCount,masalaCount;
+	
+	void setStock(int mgStock,int dStock,int oStock,int pStock,int mStock) {
+		maggieCount=mgStock;
+		dosaCount=dStock;
+		oilCount=oStock;
+		panipuriCount=pStock;
+		masalaCount=mStock;
+	}
+	void maggiPacket(int buyMaggi)
```
harshvegada, line 14: you direct deduct request item from stock, first you should check wether requested item is less or more and based upon you should perform operation.
like when ever you go to store they not always have your request stock right.
Check other's code for the same

MonikaZankar: I'll do it again..

KrishnaKTechnocredits (commented):
Merging

PR #396, opened 2021-08-11
Himanshu_Assignment_9

himanshugpt56:
image

### DIFF
--- a/src/himanshu/Assignment_9.java
+++ b/src/himanshu/Assignment_9.java
@@ -1,0 +1,68 @@
+package himanshu;
+
+public class Assignment_9 {
+
+	int maggieQuantity= 50;
+	int dosaQuantity= 43;
+	int paniPuriQuantity=43;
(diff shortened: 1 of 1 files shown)


src/himanshu/Assignment_9.java
```diff
@@ -0,0 +1,68 @@
...
+
+	int maggieQuantity= 50;
+	int dosaQuantity= 43;
+	int paniPuriQuantity=43;
+	int pouchesQuantity=39;
+	int masalasQuantity=73;
+		
+	void stocks(String itemName, int itemQuantity) {
+		if(itemName.equalsIgnoreCase("Maggie")) {
+			maggieQuantity = maggieQuantity-itemQuantity;
```
harshvegada, line 13: before deduct you should check whether do you have stock or not?
Same for all methods

KrishnaKTechnocredits (commented):
merging

PR #397, opened 2021-08-11
Santosh_Assignment_9

Santoshghige:
image


nehatawari:
image
can check condition like if(maggie>0) and then start with if and else.Also think about partial order required and not required by customer.Rest is OK

PR #404, opened 2021-08-11
Madhavi_Assignment_9

Madhavi14C:
image

### DIFF
--- a/src/madhavi/Assignment_9.java
+++ b/src/madhavi/Assignment_9.java
@@ -1,0 +1,120 @@
+package madhavi;
+
+public class Assignment_9 {
+	int totalMaggieCount;
+	int totalDosaCount;
+	int toalPouchesCount;
+	int totalMasaleCount;
+	int totalPanipuriCount;
+
+	void setAvailableStock(int maggie, int dosa, int oilpouches, int panipuri, int masale) {
+		totalMaggieCount = maggie;
+		totalDosaCount = dosa;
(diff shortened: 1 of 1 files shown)


src/madhavi/Assignment_9.java
```diff
@@ -0,0 +1,120 @@
...
+
+	void purchasedItems(int maggieUnits, int dosaUnits, int oilPouchUnits, int paniPuriUnits, int masaleUnits) {
+		System.out.println("**********Items Ordered**********");
+		System.out.println("Maggie-" + maggieUnits + "\n" + "Dosa- " + dosaUnits + "\n" + "Oil- " + oilPouchUnits + "\n"
+				+ "Panipuri- " + paniPuriUnits + "\n" + "Masala- " + masaleUnits);
+
+		// Purchase Maggie
+		if (totalMaggieCount >= maggieUnits) {
+			totalMaggieCount = totalMaggieCount - maggieUnits;
+		} else if (maggieUnits > totalMaggieCount && totalMaggieCount != 0) {
```
harshvegada, line 65: I think this second if condition never execute as first & second looks exact same.
remove 2nd one (think on this)

else condition can check on showAvailableInStockItems or showOutOfStockItems method

Here u should write 
if (totalMaggieCount < maggieUnits) then you should print available stock

harshvegada (approved):
Rest looks good!

PR #408, opened 2021-08-11
Chandani_Assignment_9

chandni0604:
image

Output after updates in program (Incorporated review comments ) 
image

### DIFF
--- a/src/chandani/Assignment9.java
+++ b/src/chandani/Assignment9.java
@@ -1,0 +1,84 @@
+package chandani;
+
+class Assignment9{
+	
+	int maggieCnt;
+	int dosaCnt;
+	int  pouchesCnt;
+	int panipuriCnt;
+	int masalaCnt;
+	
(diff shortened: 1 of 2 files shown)


a-malvi (commented):
Create a common method rather than method for all items individually.

chandni0604 (commented):
Incorporated the review comments

src/chandani/Assignment9.java
```diff
@@ -0,0 +1,84 @@
...
+		maggieCnt = maggieQty ;
+		dosaCnt = dosaQty;
+		pouchesCnt = pouchesQty;
+		panipuriCnt = panipuriQty;
+		masalaCnt = maggieQty;
+	}
+	
+	void maggieinStockorNot(String item,int maggieAmt){
+		
+		maggieCnt = maggieCnt - maggieAmt;
```
KrishnaKTechnocredits, line 21: You should check whether request item available or not, based upon you should provide it. 
Do you agree?

KrishnaKTechnocredits (approved):
Conditions require, rest code looks good to e=me
For condition check others code

PR #411, opened 2021-08-11
Apurva_Assignment_9

ApurvaD25:
image

### DIFF
--- a/src/apurva/Assignment_9.java
+++ b/src/apurva/Assignment_9.java
@@ -1,0 +1,112 @@
+package apurva;
+
+public class Assignment_9 {
+
+	static int maggie = 50;
+	static int dosa = 43;
+	static int pouches = 39;
+	static int panipuri = 43;
+	static int masala = 73;
+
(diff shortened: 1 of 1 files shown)


src/apurva/Assignment_9.java
```diff
@@ -0,0 +1,112 @@
+package apurva;
+
+public class Assignment_9 {
+
```
MadhaviKhasbage, line 4: Give meaningful name .. ex: maggieQuantiry, dosaQuantity

src/apurva/Assignment_9.java
```diff
@@ -0,0 +1,112 @@
+package apurva;
+
```
MadhaviKhasbage, line 2: Add assignment text here so you can understand requirement after 2-3 month also

MadhaviKhasbage (approved):
Code looks good to me!

MadhaviKhasbage (approved):
Code looks good to me!

PR #416, opened 2021-08-11
Ashish_Assignment_9

KrAshish2610:
image


a-malvi (approved):
Suggested changes :

line # 43 --> qtyMaggie <= 0 as quantity can be 0 in such case Maggie is out of stock.
Rest code looks good.

PR #419, opened 2021-08-12
Ashit_Assignment_9

Ashit46:
image

### DIFF
--- a/src/ashit/ShopKeeper.java
+++ b/src/ashit/ShopKeeper.java
@@ -1,0 +1,110 @@
+package ashit;
+public class ShopKeeper {
+
+	static int totalMaggiePacketQuantity;
+	static int totalDosaPacketQuantity;
+	static int totalOilPacketQuantity;
+	static int totalPanipuriPacketQuantity;
+	static int totalMasalaPacketQuantity;
+
+	void setIntialquantity(int initialMaggieQuantity, int initialDosaQuantity, int initialOilQuantiy, int initialPanipuriQuantity, int initialMasalaQuantity) {
+		totalMaggiePacketQuantity = initialMaggieQuantity;
+		totalDosaPacketQuantity = initialDosaQuantity;
+		totalOilPacketQuantity = initialOilQuantiy;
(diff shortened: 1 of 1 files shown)


src/ashit/ShopKeeper.java
```diff
@@ -0,0 +1,110 @@
...
+
+		System.out.println("**********Maggie Packets ***********");
+		if(totalMaggiePacketQuantity >= demandmaggiepacket) {
+			totalMaggiePacketQuantity = totalMaggiePacketQuantity - demandmaggiepacket;
+			System.out.println("Order deliever to the customer");
+
+		}else{
+			System.out.println("Product is out of stock");	
+		}
+
```
VaibhavTati, line 28: Statement mention in the else block is not correct.
It might be possible that 
totalMaggiePacketQuantity = 45
demandmaggiepacket = 50

Then we cant say **Product is out of stock**, it is better to write **Order not able to serve as requested quantity is greater than available quantity in stock**

src/ashit/ShopKeeper.java
```diff
@@ -0,0 +1,110 @@
...
+		System.out.println("********** Masala Packets **********");
+		if(totalMasalaPacketQuantity >= demandmasalpacket) {
+			totalMasalaPacketQuantity = totalMasalaPacketQuantity - demandmasalpacket;
+			System.out.println("Order deliever to the customer");
+
+		}else{
+			System.out.println("Product is out of stock");		
+		}
```
VaibhavTati, line 63: Statement mention in the above all else blocks in **demandProduct** method is not correct.
It might be possible that,

totalMaggiePacketQuantity = 45
demandmaggiepacket = 50

Then we cant say **Product is out of stock**, it is better to write **Order not able to serve as requested quantity is greater than available quantity in stock**

VaibhavTati:
Just to note one more thing, this can also be possible without using static variable.

KrishnaKTechnocredits (approved):
Code looks good to me!
Keep doing well!

PR #422, opened 2021-08-12
Rasika_Assignment_10

RasikaNitin:
image

### DIFF
--- a/src/rasika/Assignment10.java
+++ b/src/rasika/Assignment10.java
@@ -3,19 +3,19 @@ package rasika;
 public class Assignment10 {
 	int count = 0;
 	int sum = 0;
-	void verifyPrimeNumber(int number) {
+	void verifyPrimeNumber(int isPrimeNumber) {
 		boolean status = true;
-		for(int index = 2; index < number; index++) {
-			if(number % 2 == 0) {
+		for(int index = 2; index < isPrimeNumber; index++) {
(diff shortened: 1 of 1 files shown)


src/rasika/Assignment10.java
```diff
@@ -0,0 +1,47 @@
+package rasika;
+
+public class Assignment10 {
+	int count = 0;
+	int sum = 0;
+	void verifyPrimeNumber(int number) {
+		boolean status = true;
+		for(int index = 2; index < number; index++) {
+			if(number % 2 == 0) {
```
surabhi-9, line 9: Validation is wrong. have to use like number % index = = 0 then booleanFlag=false, then break

src/rasika/Assignment10.java
```diff
@@ -0,0 +1,47 @@
+package rasika;
+
+public class Assignment10 {
+	int count = 0;
+	int sum = 0;
+	void verifyPrimeNumber(int number) {
+		boolean status = true;
```
surabhi-9, line 7: try to use meaningfull variable name like isPrimeNumber .

surabhi-9 (changes requested):
Please rewrite the validation for prime number again with correct logic

KrishnaKTechnocredits (approved):
Code looks good to me!
Keep doing well!

PR #427, opened 2021-08-12
Neha_Assignment_10

NehaJTechnocredits:
**Output:**
image

### DIFF
--- a/src/neha/PrimeNumberAssignment10.java
+++ b/src/neha/PrimeNumberAssignment10.java
@@ -1,0 +1,41 @@
+package neha;
+
+public class PrimeNumberAssignment10 {
+	int count = 0;
+	int sum = 0;
+
+	void verifyIfNumberIsPrime(int number) {
+		boolean status = true;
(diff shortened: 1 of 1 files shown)


src/neha/PrimeNumberAssignment10.java
```diff
@@ -0,0 +1,41 @@
+package neha;
+
+public class PrimeNumberAssignment10 {
+	int count = 0;
+	int sum = 0;
+
+	void verifyIfNumberIsPrime(int number) {
+		boolean status = true;
+		for (int index = 2; index < number; index++) {
```
SandeshMolane, line 9: can use index <= number/2 for efficient time complexity. Rest looks good to me!

SandeshMolane (approved):
Code looks good to me!

src/neha/PrimeNumberAssignment10.java
NehaJTechnocredits: Thanks for suggestion. I will try to think for efficient time complexity while writing code.

PR #434, opened 2021-08-12
JagratiB_Assignment_10

Jagrati31:
image

### DIFF
--- a/src/jagrati/Assignment_10.java
+++ b/src/jagrati/Assignment_10.java
@@ -1,9 +1,10 @@
 package jagrati;
 
 public class Assignment_10 {
-	int count;
-	int sum;
-	int avg;
+	int count= 0;
+	int sum =0;
+	float avg;
+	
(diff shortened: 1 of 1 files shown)


src/jagrati/Assignment_10.java
```diff
@@ -0,0 +1,55 @@
...
+				break;
+			
+		    }
+		}
+		if (status == true) {
+			
+		System.out.println(number +" Number is prime");
+		count++;
+		sum = sum + number;
+		avg = sum/ count;
```
harshvegada, line 23: you should not calculate avg every time, you should check this at end in separate method. 
that will be easy

harshvegada (approved):
Code looks good to me,
Keep ti up!

src/jagrati/Assignment_10.java
Jagrati31: Sure ... will change my code as per your suggestion

PR #438, opened 2021-08-12
Ritabrata_Assignment10

Ritabrata1996:
image

Assignment 9 added as pull taken before "Ritabrata_Assignment9" merged

### DIFF
--- a/src/ritabrata/Assignment10.java
+++ b/src/ritabrata/Assignment10.java
@@ -1,0 +1,54 @@
+package ritabrata;
+
+public class Assignment10 {
+	double primeSum;
+	double countPrime;
+	void checkPrimeNo(int number) {
+		boolean status= true;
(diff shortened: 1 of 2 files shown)


src/ritabrata/Assignment10.java
```diff
@@ -0,0 +1,54 @@
...
+	double countPrime;
+	void checkPrimeNo(int number) {
+		boolean status= true;
+		for(int index=2; index<number; index++) {
+			if(number % index==0) {
+				status=false;
+				break;
+			}
+		}
+		if(status == true) {
```
surabhi-9, line 14: we dont need to validate status == true or not because if condition needs boolean value so if you use if(status) then it means if(true) then go inside else not go inside block

surabhi-9 (approved):
code looks good to me. please read comment on assignment 10

PR #440, opened 2021-08-12
Ashit_Assignment_10

Ashit46:
image

### DIFF
--- a/src/ashit/PrimeNumber.java
+++ b/src/ashit/PrimeNumber.java
@@ -1,0 +1,50 @@
+package ashit;
+
+public class PrimeNumber {
+	int sum;
+	int primecount;
+	int average;
+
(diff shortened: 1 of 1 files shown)


src/ashit/PrimeNumber.java
```diff
@@ -0,0 +1,50 @@
...
+				status = false;
+				System.out.println(number + " is not prime number number");
+				break;
+			}
+		}
+		if (status) {
+			System.out.println(number + " is prime number");
+			sum = sum + number;
+			primecount++;
+			average = sum / primecount;
```
KrishnaKTechnocredits, line 21: here not require to check avg all the time, you can do this in separate method

KrishnaKTechnocredits (approved):
Code looks good to me!
Think on comment

PR #448, opened 2021-08-12
Tried with Assignment 9 purchase items with multiple class

bhaktibk:
image


bhaktibk:
Sorry i have attached wrong screen shot

KrishnaKTechnocredits (approved):
This become bit complex,
you can handle this in single class

PR #456, opened 2021-08-12
Sagar_Assignment10

sagaanna:
image

### DIFF
--- a/src/sagar_Anasane/PrimeNumber.java
+++ b/src/sagar_Anasane/PrimeNumber.java
@@ -1,0 +1,50 @@
+package sagar_Anasane;
+
+class PrimeNumber{
+	int primeSum = 0;
+	int count = 0;
+	int avg = 0;
+	void verifyNumberIsPrime(int num){
+		boolean status = true;
+		for (int index=2; index < num ;index ++){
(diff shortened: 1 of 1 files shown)


src/sagar_Anasane/PrimeNumber.java
```diff
@@ -0,0 +1,50 @@
+package sagar_Anasane;
+
+class PrimeNumber{
+	int primeSum = 0;
+	int count = 0;
+	int avg = 0;
```
nehatawari, line 6: Average should be in float or double

src/sagar_Anasane/PrimeNumber.java
```diff
@@ -0,0 +1,50 @@
...
+	void verifyNumberIsPrime(int num){
+		boolean status = true;
+		for (int index=2; index < num ;index ++){
+			if (num % index == 0){
+				status = false;
+			break;
+			}
+		}
+		if (status == true){
+			System.out.println(num +"is a Prime Number");
```
nehatawari, line 16: Give spacing between number and message in println.there should be space between double quote and is

nehatawari (approved):

PR #457, opened 2021-08-12
Suchita_Assignment_10

shuchkrish:
image


Upasana1703:
Count and sum should not be static. Because they depends on range. So, they may have different values for different objects.

Upasana1703 (approved):

PR #460, opened 2021-08-12
Priti _Assingment _10

psnilange:
image


src/priti/Assingment_8.java
```diff
@@ -0,0 +1,70 @@
...
+	int currentBalance= 10000;
+	String customerName;
+	static int overAllDebitCnt;
+	static int overAllcreditCnt;
+		
+	void setCutomerDetails(String name){
+		customerName = name;
+	}
+	
+	void debitAmt(int amt){
```
Shreyash-Atal, line 14: while debiting amount, we should check that debit Amount is less than current balance else dont debit amount and show error message if required that balance is insufficient

src/priti/Assingment_8.java
```diff
@@ -0,0 +1,70 @@
...
+		currentBalance = currentBalance-amt;
```
surabhi-9, line 15: while debiting amount, validate if debit amount is less than current balance else show some message that can not debit amount as balance is insufficient if required

src/priti/Assingment_9.java
```diff
@@ -0,0 +1,102 @@
...
+	void setInitialQty(int mgQty, int dosaQty, int pouchQty, int paniQty) {
+	
+		maggie = mgQty;
+		dosa = dosaQty;
+		pouch = pouchQty;
+		panipuri = paniQty;
+	}
+	
+	void purchase1(int magQty) {
+		if (magQty>maggie) {
```
surabhi-9, line 21: you can write more validations like if currentQty <=0,  show message "out of stock" else if requestedQty <=currentQty then deduct quantity from currentqty. Rest code looks good to me

surabhi-9 (approved):
Please try to apply more validations in logic afterwards. Rest logic seems very good

PR #463, opened 2021-08-12
Suruchi_Assignment_10

suruGIT21:
image

### DIFF
--- a/src/suruchi/Assignment_10.java
+++ b/src/suruchi/Assignment_10.java
@@ -4,8 +4,8 @@ public class Assignment_10 {
 
 	int startNumber;
 	int endNumber;
-	int counter;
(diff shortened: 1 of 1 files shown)


src/suruchi/Assignment_10.java
```diff
@@ -0,0 +1,63 @@
+package suruchi;
+
+public class Assignment_10 {
+
+	int startNumber;
+	int endNumber;
+	static int counter;
```
harshvegada, line 7: non-static also work for counter & sum

harshvegada (approved):
Code looks good to me,
 Good to merge!

src/suruchi/Assignment_10.java
suruGIT21: okay

PR #464, opened 2021-08-12
bharat_assinment_10

bharattanpure:
prime number

### DIFF
--- a/src/bharat/assignment/Assignment_11.java
+++ b/src/bharat/assignment/Assignment_11.java
@@ -1,0 +1,54 @@
+package bharat.assignment;
+
+public class Assignment_11 {
+
+	int sum;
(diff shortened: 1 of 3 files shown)


src/bharat/assignment/Assignment_11.java
```diff
@@ -0,0 +1,54 @@
...
+	
+	void getPrimeNumber(int value) {
+		int index = 1;
+		while(cntPrime < value) {
+			primeNumber(index);
+			index++;
+		}
+	}
+	
+	void sumOf() {
```
KrishnaKTechnocredits, line 35: method should be parametrised

KrishnaKTechnocredits (approved):
Code looks good

PR #465, opened 2021-08-12
SiddharthT_assignment_10

siddharth-tambe:
image


Upasana1703 (approved):
Current average value is not correct. So, use double or float data type for average. Also try to use boolean data type for status.

PR #466, opened 2021-08-12
Sameer_Assignment_10

SZalke:
image

### DIFF
--- a/src/sameer/Assignment10.java
+++ b/src/sameer/Assignment10.java
@@ -1,0 +1,50 @@
+package sameer;
+
+public class Assignment10 {
+	float count;
+	float sum;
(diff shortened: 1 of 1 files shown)


src/sameer/Assignment10.java
```diff
@@ -0,0 +1,50 @@
+package sameer;
+
+public class Assignment10 {
+	float count;
+	float sum;
+
+	void verifyPrimeNumber(int number) {
+		boolean status = true;
+		for (int index = 2; index < number; index++) {
```
nehatawari, line 9: index< number/2 optimization.

nehatawari (approved):

PR #470, opened 2021-08-12
Astha_Assignment_10

AsthaPrakash:
image


src/astha/Assignment10/PrimeNumber.java
```diff
@@ -0,0 +1,40 @@
+package astha.Assignment10;
+
```
MadhaviKhasbage, line 2: Add Assignment text here so it will help you to understand requirement after 2-3 months also

MadhaviKhasbage (approved):
Code looks good to me!

PR #475, opened 2021-08-12
Monika_Assignment_10

MonikaZankar:
image

### DIFF
--- a/src/monika/Assignment_10.java
+++ b/src/monika/Assignment_10.java
@@ -1,0 +1,56 @@
+package monika;
+
+public class Assignment_10 {
+	
+static int count=0;
+static int sum=0;
+static float average=0;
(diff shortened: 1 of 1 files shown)


src/monika/Assignment_10.java
```diff
@@ -0,0 +1,56 @@
...
+			if(number%index==0) {
+				flag=false;
+				break;
+			}		
+		}
+		if(flag) {
+			System.out.println(number+" is Prime Number");
+			count=count+1;
+			sum=sum+number;
+			average=sum/count;
```
harshvegada, line 21: every time you are doing sum & finding avg, you should create a method which tell at the end what is the avg of this.

harshvegada (approved):
take care of comment & think on such repeated scenario how you can reduce those.

PR #484, opened 2021-08-12
Avinash_Assignment_10

AvinashSPatil:
image


src/avinash/PrimeNumber.java
```diff
@@ -0,0 +1,60 @@
+package avinash;
+
```
MadhaviKhasbage, line 2: Add assignment text here, so you can understand the requirement after 2-3 month also

MadhaviKhasbage (approved):
Code looks good to me!

PR #487, opened 2021-08-12
Bhagyashree_Assignment10

BhagyashreeJYeole:
image

### DIFF
--- a/src/bhagyashree/Assignment10.java
+++ b/src/bhagyashree/Assignment10.java
@@ -3,7 +3,7 @@ package bhagyashree;
 public class Assignment10 {
 	int count = 0;
 	int sum = 0;
-	int average;
+	int average=0;
 
 	void isPrimeNumber(int number) {
 		boolean status = true;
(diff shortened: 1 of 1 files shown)


src/bhagyashree/Assignment10.java
```diff
@@ -0,0 +1,56 @@
+package bhagyashree;
+
```
MadhaviKhasbage, line 2: Add assignment text here, so you can understand the requirement after 2-3 month also

src/bhagyashree/Assignment10.java
```diff
@@ -0,0 +1,56 @@
...
+	void isPrimeNumber(int number) {
+		boolean status = true;
+		for (int index = 2; index < number; index++) {
+			if (number % index == 0) {
+				status = false;
+				System.out.println(number + " is not prime");
+				break;
+			}
+		}
```
MadhaviKhasbage, line 16: Program out put is wrong,
you should have written below logic in open and close curly bracket. your logic get changed coz you did not add brackets. Add brackets and check.

if (status == true){
		System.out.println(number + " is prime");
		count++;
		sum = sum + number;
		average = sum / number;
}

BhagyashreeJYeole:
Done with changes
image

MadhaviKhasbage (approved):
now code looks good to me!!

PR #494, opened 2021-08-12
Riten_Assignment_10

ritenthakkar:
image


surabhi-9 (approved):
code looks good to me. just an suggestion , its better to write generic method and other calculation like avg, sum should be done in another method, so that any one should be able to access verifyPrimeMethod to check if num is prime or not

PR #495, opened 2021-08-12
Shraddha_Assignment_10

shraddhaatal:
image


nehatawari:
Optimize way is for(int index = 2;index < number/2;index++) 
Average should be in float or double

PR #497, opened 2021-08-12
ShwetaD_Assignment10

ShwetaDharmadhikari:
image


src/shweta_Dharmadhikari/PrimeNumber/Assignment10.java
```diff
@@ -0,0 +1,54 @@
+package shweta_Dharmadhikari.PrimeNumber;
+
+public class Assignment10 {
+	int count;
+	int sum;
+	int average;
```
Upasana1703, line 6: Use float or double to store the average. So, that decimal part can be included.

Upasana1703 (approved):

PR #504, opened 2021-08-13
Shyam_Assignment_10

shyamanasane:
image

### DIFF
--- a/src/shyam/Shyam_Assignment_10.java
+++ b/src/shyam/Shyam_Assignment_10.java
@@ -1,0 +1,47 @@
+package shyam;
+
+public class Shyam_Assignment_10 {
+	int count;
+	int sum;
+	int avg;
+
+	void verifyNumberIsPrime(int number) {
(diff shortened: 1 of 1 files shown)


src/shyam/Shyam_Assignment_10.java
```diff
@@ -0,0 +1,47 @@
+package shyam;
+
+public class Shyam_Assignment_10 {
+	int count;
+	int sum;
+	int avg;
```
Upasana1703, line 6: Use float or double data type to store average value. Because in int decimal part will be not included.

Upasana1703 (approved):

src/shyam/Shyam_Assignment_10.java
KrishnaKTechnocredits: Yes, true

KrishnaKTechnocredits (approved):
Code Logic looks good to me!
work on comments

PR #505, opened 2021-08-13
Shyam_Assignment_9

shyamanasane:
image

### DIFF
--- a/src/shyam/ShoppingSystem.java
+++ b/src/shyam/ShoppingSystem.java
@@ -1,0 +1,102 @@
+package shyam;
+
+public class ShoppingSystem {
+	static int maggieQuantity = 50;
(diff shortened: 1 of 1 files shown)


src/shyam/ShoppingSystem.java
```diff
@@ -0,0 +1,102 @@
+package shyam;
+
+public class ShoppingSystem {
+	static int maggieQuantity = 50;
+	static int dosaQuantity=43;
+	static int pouchesQuantity=39;
+	static int panipuriQuantity=43;
+	static int masalaQuantity=73;
+	
+	static void availableStock() {
```
Upasana1703, line 10: According to question we should only print items which are in stock.

Upasana1703 (approved):
Use meaningful variable names.

PR #512, opened 2021-08-13
Monika_Assignment_11

MonikaZankar:
image

### DIFF
--- a/src/monika/Assignment_11.java
+++ b/src/monika/Assignment_11.java
@@ -1,0 +1,55 @@
+package monika;
+
+public class Assignment_11 {
+
+	int count = 0;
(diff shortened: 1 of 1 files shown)


src/monika/Assignment_11.java
```diff
@@ -0,0 +1,55 @@
...
+		return flag;
+			
+	}
+	void requiredPrimeNumbers(int requiredCount) {
+		int number = 2;
+		System.out.print("First " + requiredCount + " Prime Numbers are: ");
+		while (count < requiredCount) {
+			if(verifyNumberisPrime(number)) {
+				System.out.print(number + " ");
+			count++;
```
harshvegada, line 28: I doubt about this line
this should be outside the if block

PR #527, opened 2021-08-13
Assignment_11

amolpawar4334:
image

### DIFF
--- a/src/amol_Pawar/Assignment_11.java
+++ b/src/amol_Pawar/Assignment_11.java
@@ -1,0 +1,64 @@
+package amol_Pawar;
+
+public class Assignment_11 {
+	int count = 0;
+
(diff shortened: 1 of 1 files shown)


src/amol_Pawar/Assignment_11.java
```diff
@@ -0,0 +1,64 @@
...
+			getPrimeNum(index);
+			index++;
+		}
+	}
+
+	void sumOfNumbers() {
+		int sum = 0;
+		int count = 0;
+		int index = 1;
+		while (sum < 100) {
```
KrishnaKTechnocredits, line 36: method should be parameterised

KrishnaKTechnocredits (approved):
Code looks good to me!
Keep doing well!

PR #528, opened 2021-08-13
SomnathK_Assignment_11

SamWaltan:
image

### DIFF
--- a/src/somnath/Assignment11.java
+++ b/src/somnath/Assignment11.java
@@ -1,0 +1,57 @@
+
+//Assignment11-Whileloop
+
+package somnath;
+public class Assignment11 {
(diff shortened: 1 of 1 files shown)


src/somnath/Assignment11.java
```diff
@@ -0,0 +1,57 @@
...
+
+	Boolean verifyPrimeNumber(int number) {
+		Boolean status = true;
+		for (int index = 2; index <= number / 2; index++) {
+			if (number % index == 0) {
+				status = false;
+				break;
+			}
+		}
+		if (status)
```
Upasana1703, line 28: No need of if else block. Directly return the status.

Upasana1703 (approved):
Code looks good to me.

PR #545, opened 2021-08-13
Sayaliv assignment  11

sayalivt:
image

### DIFF
--- a/src/sayali_V/Assignment_11.java
+++ b/src/sayali_V/Assignment_11.java
@@ -1,0 +1,65 @@
+package sayali_V;
+
+public class Assignment_11 
+{
+	boolean verifyPrimeNumber(int number)
+	{
+		boolean status=true;
+		int index=2;
+		while(index<number)
+		{
+			if(number%index==0)
+			{
+				status = false;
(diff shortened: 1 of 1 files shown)


src/sayali_V/Assignment_11.java
```diff
@@ -0,0 +1,65 @@
...
+			if(number%index==0)
+			{
+				status = false;
+				break;	
+			}
+			index++;
+		}
+		if(status) 
+		{
+			System.out.print(number+ " , ");
```
nehatawari, line 20: better not give , and use println as after last number also , is coming

src/sayali_V/Assignment_11.java
```diff
@@ -0,0 +1,65 @@
...
+		int num1=1;
+		int sum=0;
+		int totalcount=0;
+		while(sum<range)
+		{
+			sum=sum+num1;
+			num1=num1+1;
+			totalcount++;
+		}
+		System.out.print(totalcount);
```
nehatawari, line 50: give proper message

nehatawari (approved):

PR #554, opened 2021-08-13
Ramkrishna assignment 11

ramunikam:
Screenshot 2021-08-13 at 11 33 32 PM

### DIFF
--- a/src/ramkrishna/PrimeNumber.java
+++ b/src/ramkrishna/PrimeNumber.java
@@ -1,0 +1,61 @@
+package ramkrishna;
+
+public class PrimeNumber {
+
+	int count = 0;
+	int sum = 0;
+	float avg;
+
+	void numberIsPrime(int number) {
+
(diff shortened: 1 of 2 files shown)


src/ramkrishna/PrimeNumber.java
```diff
@@ -0,0 +1,61 @@
...
+	void numberIsPrime(int number) {
+
+		boolean status = true;
+		for (int index = 2; index < number; index++) {
+			if (number % index == 0) {
+				status = false;
+				break;
+			}
+		}
+		if (status == true) {
```
surabhi-9, line 18: no need to check if status boolean variable is true or not because its self has value true or false. hope understand

surabhi-9 (approved):
code logic is good but can improved if you  can make some generic method which can return boolean or int based on which we can deside its prime or not. so that resuability of code can be achieved

PR #565, opened 2021-08-13
SayliVyavhare_Assignment_10

Saylivyavhare:
image

### DIFF
--- a/src/sayli_Vyavhare/Assignment10_Prime_Sum.java
+++ b/src/sayli_Vyavhare/Assignment10_Prime_Sum.java
@@ -1,0 +1,38 @@
+package sayli_Vyavhare;
+public class Assignment10_Prime_Sum {
+	int  totalPrime=0,averagePrime=0;
+	int countPrime=0;
+	
+	void primeOrNot(int number) {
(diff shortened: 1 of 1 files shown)


src/sayli_Vyavhare/Assignment10_Prime_Sum.java
```diff
@@ -0,0 +1,38 @@
+package sayli_Vyavhare;
+public class Assignment10_Prime_Sum {
+	int  totalPrime=0,averagePrime=0;
```
nehatawari, line 3: Average should be in float

src/sayli_Vyavhare/Assignment10_Prime_Sum.java
```diff
@@ -0,0 +1,38 @@
...
+	int countPrime=0;
+	
+	void primeOrNot(int number) {
+		boolean flag=true;
+		
+		for(int index=2;index<number;index++) {
```
nehatawari, line 9: for(int index=2;index<=number/2;index++)
optimize way

nehatawari (approved):

PR #574, opened 2021-08-13
Purshottam_Assignment11

purshottam94:
image

### DIFF
--- a/src/purshottamJoshi/Assignment11.java
+++ b/src/purshottamJoshi/Assignment11.java
@@ -1,0 +1,46 @@
+package purshottamJoshi;
+
+public class Assignment11 {
+	int countNumbers = 0;
+
+	void printNPrimeNumbers(int number) {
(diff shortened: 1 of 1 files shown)


src/purshottamJoshi/Assignment11.java
```diff
@@ -0,0 +1,46 @@
...
+			flag = true;
+			int index = num - 1 ;
+			while(index >1) {
+				if (num % index == 0) {
+					flag = false;
+					break;
+				}
+				index--;
+			}
+			if (flag == true) {
```
surabhi-9, line 20: dont need to check if flag is true or not, if needs boolean value to go inside its block that is flag can store value true or false hence we dont need to check again if its true or not

surabhi-9 (approved):
Prime logic is running correctly but try to simplify more,  as its seems like complex logic. rest code look good to me

PR #584, opened 2021-08-14
Aniket_Assignment_10

aniket0811:
image

### DIFF
--- a/src/aniket/PrimeNumbers.java
+++ b/src/aniket/PrimeNumbers.java
@@ -1,0 +1,61 @@
+package aniket;
+
+public class PrimeNumbers {
+	int count = 0;
+	int sum =0;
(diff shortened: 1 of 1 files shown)


src/aniket/PrimeNumbers.java
```diff
@@ -0,0 +1,61 @@
...
+				status = false;
+				break;
+			}
+
+		}
+		if (status) {
+			System.out.println(number + " is prime number");
+			count++;
+			sum=sum+number;
+			avg=(float)sum/count;
```
KrishnaKTechnocredits, line 22: here not require to check avg all the time, you can do this in separate method

KrishnaKTechnocredits (approved):
Code matches expected requirement!

PR #606, opened 2021-08-14
Varun_Assignment_12

varunj822:
image

### DIFF
--- a/src/varun/Mall.java
+++ b/src/varun/Mall.java
@@ -1,0 +1,72 @@
+package varun;
+
+public class Mall {
+	String customerName;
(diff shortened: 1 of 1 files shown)


src/varun/Mall.java
```diff
@@ -0,0 +1,72 @@
+package varun;
+
+public class Mall {
+	String customerName;
+	String LastName;
```
harshvegada, line 5: Variable name should be in CamelCase and it should starts with small letters

harshvegada (approved):
Code logic looks good to me!

PR #608, opened 2021-08-14
Neha_Assignment_12

NehaJTechnocredits:
**Output:**
image

### DIFF
--- a/src/neha/MallAssignment12.java
+++ b/src/neha/MallAssignment12.java
@@ -1,0 +1,82 @@
+package neha;
+
+public class MallAssignment12 {
+	String customerName;
+	Boolean successfulPayment = false;
+
+	int itemAddedToCart(int itemCount) {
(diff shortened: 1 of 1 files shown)


src/neha/MallAssignment12.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	void displayCutomerInfo(String section, int itemCount, String paymentMode) {
+		System.out.println(customerName + " bought " + itemCount + " items and payment done using " + paymentMode
+				+ " in section " + section);
+	}
+
+	void needBill(Boolean flag) {
+		if (flag&&successfulPayment)
```
harshvegada, line 38: with && condition you have putted successfulPayment(have false value so it will never go inside If block)

harshvegada (approved):
Few extra stuff, but Code logic looks good to me!

PR #610, opened 2021-08-14
Riten_Assignment_11

ritenthakkar:


src/riten/PrimeWhileCheck.java
```diff
@@ -0,0 +1,74 @@
...
+				// System.out.println(number+" Num is not Prime");
+				status = false;
+				break;
+			}
+			index2++;
+		}
+		if (status) {
+			// System.out.println(number + " Num is Prime Number");
+
+			boolean flag = true;
```
surabhi-9, line 24: no flag variable and if condition needed, directly we can check if(status), check if(sum<=range) then call sumPrime method.

surabhi-9 (approved):
rest code looks good to me

src/riten/PrimeWhileCheck.java
ritenthakkar: Yes Yes. I will make changes in code and commit the same. you can then please merger the code to master.

surabhi-9: its merged already. just added comment so that you can consider in further programming. :)

PR #618, opened 2021-08-14
ShwetaD_Assignment12

ShwetaDharmadhikari:
image

### DIFF
--- a/src/shweta_Dharmadhikari/Mall.java
+++ b/src/shweta_Dharmadhikari/Mall.java
@@ -1,0 +1,59 @@
+package shweta_Dharmadhikari;
+
+public class Mall {
+
+	int itemAddedToCart(int itemCount) {
+		return itemCount;
+	}
+	
+	String customerName(String name) {
+		return name;
+	}
+	
+	String paymentDoneUsing(String paymentMethod) {
+		//payMethod = paymentMethod;
+		return paymentMethod;
+	}
+	
+	void displayCutomerInfo(String name, int item,String payMethod,String sectionName) {
+		System.out.println(name+" brought "+item+" items and payment done using "+payMethod+" in section "+sectionName);
+	}
+	
+	boolean needBill(boolean isBillRequired) {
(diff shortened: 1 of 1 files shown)


src/shweta_Dharmadhikari/Mall.java
```diff
@@ -0,0 +1,59 @@
...
+	String paymentDoneUsing(String paymentMethod) {
+		//payMethod = paymentMethod;
+		return paymentMethod;
+	}
+	
+	void displayCutomerInfo(String name, int item,String payMethod,String sectionName) {
+		System.out.println(name+" brought "+item+" items and payment done using "+payMethod+" in section "+sectionName);
+	}
+	
+	boolean needBill(boolean isBillRequired) {
```
Upasana1703, line 22: If else block not required, return isBillRequired.

src/shweta_Dharmadhikari/Mall.java
```diff
@@ -0,0 +1,59 @@
...
+		if (isBillRequired)
+			return true;
+		else
+			return false;
+	}
+	
+	boolean haveCarryBag(boolean isCarryBagRequired) {
```
Upasana1703, line 29: If else block not required, return isCarryBagRequired.

src/shweta_Dharmadhikari/Mall.java
```diff
@@ -0,0 +1,59 @@
...
+		if (isCarryBagRequired)
+			return true;
+		else
+			return false;
+	}
+	
+	String decideSection(int item,String payMethod) {
```
Upasana1703, line 36: For item=10, section will be blank.

Upasana1703 (approved):

src/shweta_Dharmadhikari/Mall.java
ShwetaDharmadhikari: nope , customer will send true or false from main method so we have to set isCarryBagRequired variable

PR #660, opened 2021-08-15
Suchita_Assignment_13

shuchkrish:
image

image

image

### DIFF
--- a/src/suchita/Program_1.java
+++ b/src/suchita/Program_1.java
@@ -1,0 +1,22 @@
+package suchita;
+
+ class Program_1 {
+	
+	int countOfPositiveNumbers(int[] number) {
+		int count = 0;
+		int num[]=number;
+		for(int index=0;index<num.length;index++){
+			if(num[index] > 0)
+			
+			count++;
+		}
+		return(count);
+	}
+	
+	public static void main(String[] args) {
+		Program_1 assignment = new Program_1();
+		int[] input = {1,-11,-44,23,55,-99,-23,-22};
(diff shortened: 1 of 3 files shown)


src/suchita/Program_1.java
```diff
@@ -0,0 +1,22 @@
+package suchita;
+
+ class Program_1 {
+	
+	int countOfPositiveNumbers(int[] number) {
+		int count = 0;
+		int num[]=number;
```
Upasana1703, line 7: no need of new variable. You can directly use number variable.

src/suchita/Program_2.java
```diff
@@ -0,0 +1,21 @@
+package suchita;
+
+ class Program_2 {
+
+	 int countOfZeros(int[] number) {
+			int count = 0;
+			int num[]=number;
```
Upasana1703, line 7: no need of new variable. You can directly use number variable.

src/suchita/Program_3.java
```diff
@@ -0,0 +1,28 @@
+package suchita;
+
+ class Program_3 {
+	 
+	int[] smallerNumbersOfGivenRange(int num) {
+		int input[] = new int[5];
+		int count=0;
+		//int num=number;
+		for(int index=num;index>num-5;num--){
+			if(count != 5){
```
Upasana1703, line 10: By using index>num-5, you are sure that loop will run only five times. So, there is no need of count variable and if check.

Upasana1703 (approved):
Code looks good to me.

PR #664, opened 2021-08-15
Ritabrata_Assignment13

Ritabrata1996:
program 1 : return count of positive numbers from given array:
image
program 2 : count number of zero's in given array:
image
program 3 : write a method which returns 5 smaller numbers of the given number:
image

### DIFF
--- a/src/ritabrata/CountZero.java
+++ b/src/ritabrata/CountZero.java
@@ -1,0 +1,38 @@
+/*Assignment 13 : 15th Aug'2021
+program 2 : count number of zero's in given array.
+int[] input = {1,-11,0,0,55,0,-23,0};
+
+output : 4
+ */
+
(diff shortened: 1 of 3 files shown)


src/ritabrata/CountZero.java
```diff
@@ -0,0 +1,38 @@
...
+				count++;
+			}
+		}
+		System.out.println("The total no if positive integer is: "+count);
+		return count;
+	}
+		
+	public static void main(String[] args) {
+		CountZero countZero= new CountZero();
+		int[] arr= new int[8];
```
harshvegada, line 24: you can write array like
int a[] = {2,6,5,6,7,87} just example.

harshvegada (approved):
Code Logic looks good to me!

PR #666, opened 2021-08-15
akankshak assignment 11

akankshaz:
image


src/akanksha_K/Assignment11.java
```diff
@@ -0,0 +1,54 @@
...
+			else {
+				System.out.println(num +" is not a prime number");
+			}
+			num++;
+	     }
+		System.out.println("");
+	}
+	
+	//program 2
+	void sum() {
```
harshvegada, line 34: Ideally we should pass param in method

harshvegada (approved):
Code Logic looks good to me!

PR #667, opened 2021-08-15
Shantanu_Assignment_13

ShantanuGTechnocredits:
image
image
image


nehatawari:
if(count >= 5)
break;
no need of this condition as array size is 5 so loop will run upto 5.

PR #676, opened 2021-08-15
BhaktiK Assignment_13

bhaktibk:
image

### DIFF
--- a/src/bhakti/Assignment_13_array.java
+++ b/src/bhakti/Assignment_13_array.java
@@ -1,0 +1,76 @@
+//Assignment 13 : 15th Aug'2021
+//
+//program 1 : return count of positive numbers from given array.
+//int[] input = {1,-11,-44,23,55,-99,-23,-22};
+//output : 3
+//
+//program 2 : count number of zero's in given array.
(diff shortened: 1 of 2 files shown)


src/bhakti/Assignment_13_array.java
```diff
@@ -0,0 +1,76 @@
...
+		int count=0;
+		for (int index=0; index<array.length; index++ ) {
+			if (array[index] ==0 ) {
+				count++;
+			}	
+		}
+		return count;
+	}
+	
+	//program 3 : write a method which returns 5 smaller numbers of the given number.
```
MadhaviKhasbage, line 45: try like this >>

void get5smallerNumber(int num) {
			int[] numers= new int[5];
			String strOutPut="";

			for (int index=0; index <5; index++) {
				num--;
				numers[index]= num;
				strOutPut += numers[index]+",";
			}
			System.out.println(strOutPut.substring(0, strOutPut.length()-1));
		}

MadhaviKhasbage (commented):
Code looks good to me!!
just make changes in your code as per suggestion.

PR #696, opened 2021-08-15
Akshay_Assignment_13

akshay0795:
image
image
image

### DIFF
--- a/src/akshay/Program_3.java
+++ b/src/akshay/Program_3.java
@@ -1,0 +1,25 @@
+package akshay;
+
+public class Program_3 {
+	
+	int[] getArrayofFiveNumbers(int num) {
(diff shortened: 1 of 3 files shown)


src/akshay/Program_3.java
```diff
@@ -0,0 +1,25 @@
+package akshay;
+
+public class Program_3 {
+	
+	int[] getArrayofFiveNumbers(int num) {
+		
+		int count = 0;
+		int[] input = new int[5];
+		for(int index = num; index > num- 5; index--) {
```
harshvegada, line 9: instead of num-5 in condition u can use 0

harshvegada (approved):
Code Logic looks good to me!

PR #703, opened 2021-08-15
Tanmoy roy assignment13

tanmoyroy96:
Output Program 1 to find the Positive numbers count of an Array:
image

Output Program 2 to find the Zero numbers count of an Array:
image

Output Program 3 to write a method which returns 5 smaller numbers of the given number.
image


src/tanmoy_Roy/Assignment13_1.java
```diff
@@ -0,0 +1,43 @@
...
+		for(int i=0;i<N;i++)
+			str=str+String.valueOf(array[i])+" ";
+		return "{"+str+ "}";
+	}
+	
+	public static void main(String[] args) {
+		Assignment13_1 assignment13=new Assignment13_1();
+		int[] array1= {1,-11,-44,23};
+		
+		System.out.println(assignment13.arryToSinglelineString(array1,4)+" array has " +assignment13.countPositiveIntegers(array1,4)+
```
harshvegada, line 32: u don't need to pass length of array, use property of array, & also follow best practices

src/tanmoy_Roy/Assignment13_2.java
```diff
@@ -0,0 +1,28 @@
+package tanmoy_Roy;
+
+import tanmoy_Roy.Assignment13_1;
+
+public class Assignment13_2 {
+	
+	int countZeros(int[] array,int N) {				//Same code as countPositiveIntegers() comparison to b made ==0
```
harshvegada, line 7: same as above

harshvegada (approved):
Follow best practices,
Code Logic looks good to me!

PR #704, opened 2021-08-15
Assignment13_P1

devendrachandankhede:
image
image
image


a-malvi (approved):
Code looks good.

For Assignment13_P3:
Suggestion : Rather than creating an output array again, you can directly iterate your array created in
arrayOfNumber method.

Rest all good, merged

PR #709, opened 2021-08-15
Tanmoy roy assignment14

tanmoyroy96:
Output to find the max element value:
image

Output to find the min element value:
image

Output to find the difference between max and min element value:
image


src/tanmoy_Roy/Assignment14_2.java
```diff
@@ -0,0 +1,39 @@
...
+		if(N%2==1 && array[N-1]<=min)					//explicitly handling array[N-1] for odd elements only
+			min=array[N-1];
+		return min;
+	}
+	
+	String arryToSinglelineString(int[] array,int N) {				//Method to convert array to single line string
+		String str=""; 
+		for(int i=0;i<N;i++)
+			str=str+String.valueOf(array[i])+" ";
+		return "{"+str+ "}";
```
harshvegada, line 25: Do not return value in {} format simple return value, if you have specific requirements then do it but it should be at where u capture return value.

src/tanmoy_Roy/Assignment14_3.java
```diff
@@ -0,0 +1,33 @@
...
+		Assignment14_2 assignment14_2= new Assignment14_2();
+		 
+		return assignment14_1.getMaxElement(array, N)-assignment14_2.getMinElement(array, N);
+	}
+	
+	String arryToSinglelineString(int[] array,int N) {				//Method to convert array to single line string
+		String str=""; 
+		for(int i=0;i<N;i++)
+			str=str+String.valueOf(array[i])+" ";
+		return "{"+str+ "}";
```
harshvegada, line 18: same as above

harshvegada (approved):
Code Logic looks good to me!

PR #728, opened 2021-08-16
Shantanu_Assignment_15

ShantanuGTechnocredits:
image
image
image

### DIFF
--- a/src/shantanu/assignment15/MiddleChar.java
+++ b/src/shantanu/assignment15/MiddleChar.java
@@ -1,0 +1,28 @@
+package shantanu.assignment15;
+
+public class MiddleChar {
+	char[] findMiddleCharacter(String[] input) {
+		char[] arr = new char[3];
+		for(int index = 0; index < input.length; index++) {
+			arr[index] = input[index].charAt(input[index].length()/2);
+		}
+		return arr;
+	}
+	
+	char[] createArray() {
+		String[] arr= {"hello","techno","credits"};
+		return findMiddleCharacter(arr);
+	}
+	
+	void displayMiddleCharacter() {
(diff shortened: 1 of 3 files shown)


src/shantanu/assignment15/MiddleChar.java
```diff
@@ -0,0 +1,28 @@
+package shantanu.assignment15;
```
nehatawari, line 1: please write program statement so it will be easy for you in future for reference

src/shantanu/assignment15/MiddleChar.java
```diff
@@ -0,0 +1,28 @@
...
+	char[] createArray() {
+		String[] arr= {"hello","techno","credits"};
+		return findMiddleCharacter(arr);
+	}
+	
+	void displayMiddleCharacter() {
+		char[] arr = createArray();
+		System.out.println("Middle character:");
+		for(int index = 0; index < arr.length; index++)
+			System.out.println(arr[index]);
```
nehatawari, line 21: You can print Whole string and then its middle character for better understanding.

src/shantanu/assignment15/MiddleChar.java
```diff
@@ -0,0 +1,28 @@
...
+public class MiddleChar {
+	char[] findMiddleCharacter(String[] input) {
+		char[] arr = new char[3];
+		for(int index = 0; index < input.length; index++) {
+			arr[index] = input[index].charAt(input[index].length()/2);
+		}
+		return arr;
+	}
+	
+	char[] createArray() {
```
nehatawari, line 12: You can initialize array in main method also.no need to create separate method for that

nehatawari (approved):

PR #730, opened 2021-08-16
Rushikesh Assignment 13

rushikeshjunewal:
image
image
image

### DIFF
--- a/src/rushikesh/assignment13/CountPositive.java
+++ b/src/rushikesh/assignment13/CountPositive.java
@@ -1,0 +1,28 @@
+package rushikesh.assignment13;
+
+public class CountPositive {
+	int positiveCount=0;
+	
+	int[] input = {1,-11,-44,23,55,-99,-23,-22};
+	
+	int processData() {
+		int ans=returnCountOfPositive(input);
+		return ans;
+	}
+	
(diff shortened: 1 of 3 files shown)


src/rushikesh/assignment13/CountPositive.java
```diff
@@ -0,0 +1,28 @@
+package rushikesh.assignment13;
+
+public class CountPositive {
+	int positiveCount=0;
```
surabhi-9, line 4: try to make all variables local if use of that variable is limited to that block only.

src/rushikesh/assignment13/FiveSmallNum.java
```diff
@@ -0,0 +1,20 @@
...
+		for(int index=0;index<howManySmall;index++) {
+			givenNum--;
+			output[index] = givenNum;
+		}
+		return output;
+	}
+	public static void main(String[] args) {
+		FiveSmallNum fiveSmallerNum=new FiveSmallNum();
+		int arr[] = fiveSmallerNum.getNumbers(50,5);
+		for(int i=0; i<5;i++)
```
surabhi-9, line 15: please try to use index instead of i, and use arr.length attribute in loop condition

surabhi-9 (approved):

PR #731, opened 2021-08-16
Rushikesh Assignment 14

rushikeshjunewal:
image
image
image
image

### DIFF
--- a/src/rushikesh/assignment14/DiffOddEven.java
+++ b/src/rushikesh/assignment14/DiffOddEven.java
@@ -1,0 +1,33 @@
+package rushikesh.assignment14;
+
+public class DiffOddEven {
+
+	int getDiff(int[] input) {
+	int maxOdd=input[0];
+	int maxEven=input[0];
+	
+	for(int index=0;index<input.length;index++) {
+		if(input[index]%2==0) {
+			if(input[index]>maxEven) 
+				maxEven=input[index];
+			
+		}else {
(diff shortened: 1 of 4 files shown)


surabhi-9 (changes requested):
you hv considered input[0] as even as well as odd, hence odd count and even count will miss this value, please correct the logic and push again

src/rushikesh/assignment14/DiffOddEven.java
```diff
@@ -0,0 +1,33 @@
+package rushikesh.assignment14;
+
+public class DiffOddEven {
+
+	int getDiff(int[] input) {
+	int maxOdd=input[0];
+	int maxEven=input[0];
```
surabhi-9, line 7: this logic seems incorrect, please correct and use input[0] as even or odd once at a time, both cant be even and odd

src/rushikesh/assignment14/FindMax.java
```diff
@@ -0,0 +1,25 @@
+package rushikesh.assignment14;
+
+public class FindMax {
+	
+
+	
+	int arr(int[] input) {
+		int maxNum=input[0];
+		for(int index=0;index<input.length;index++) {
```
surabhi-9, line 9: we already taken input[0] as maxNum, better to check all validation in loop from index 1 and onwards i.e for(int index=1; index<input.length; index++)

surabhi-9 (approved):

PR #734, opened 2021-08-16
Sagar_Assignment13

sagaanna:
image

### DIFF
--- a/src/sagar_Anasane/Assignment13.java
+++ b/src/sagar_Anasane/Assignment13.java
@@ -1,0 +1,44 @@
+package sagar_Anasane;
+
+class Assignment13{ 
+	int[] input1 = {1,-11,-44,23,55,-99,-23,-22};
+	int[] input2 = {1,-11,0,0,55,0,-23,0};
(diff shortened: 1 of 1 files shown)


src/sagar_Anasane/Assignment13.java
```diff
@@ -0,0 +1,44 @@
+package sagar_Anasane;
+
```
nehatawari, line 2: write down problem statement for future reference

src/sagar_Anasane/Assignment13.java
```diff
@@ -0,0 +1,44 @@
...
+				cnt = cnt+1;
+		}
+		System.out.println(cnt);
+	}
+	
+	int fiveSmallerNumber(int num){
+		int [] Number = new int[5];
+		for (int index = 0; index <5; index++){
+			num --;
+			System.out.println(num);
```
nehatawari, line 29: try to print it in main method by returning array of integer.

nehatawari (approved):

PR #763, opened 2021-08-16
Manaswi_Assignment_14_Program_1to3

ManaswiHumbe:
program 1:
image
Program 2:
image
Program 3:
image

### DIFF
--- a/src/manaswi/Assignment_14_Program3.java
+++ b/src/manaswi/Assignment_14_Program3.java
@@ -1,0 +1,36 @@
+package manaswi;
+public class Assignment_14_Program3 {
+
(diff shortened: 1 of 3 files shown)


src/manaswi/Assignment_14_Program3.java
```diff
@@ -0,0 +1,36 @@
...
+		int max = 0;
+		for (int index = 0; index < arr.length; index++) {
+			if (arr[index] > max) {
+				max = arr[index];
+			}
+		}
+		return max;
+	}
+
+	int getMinNumber(int[] arr) {
```
harshvegada, line 14: you can handle min n max in same method as well, think on this

PR #764, opened 2021-08-16
Ritabrata assignment15

Ritabrata1996:
1. program 1*:  write a program that will return difference of even length name and odd length name.
image
2. program 2 : return middle char of all elements of the String array.
image
3.program 3 : return difference between sum of even number - sum of odd numbers. Difference has to be positive. print the answer in main method.
image

### DIFF
--- a/src/ritabrata/DiffBetnMaxxOddMaxEven.java
+++ b/src/ritabrata/DiffBetnMaxxOddMaxEven.java
@@ -1,0 +1,49 @@
+/*Assignment14:15th Aug'2021
+ program 4 : print difference between max odd number and max even number from given array. difference should be positive
+
+int[] input = {1,11,44,23,55,99,23,22};
+hint : 99 - 44
+output : 55
+ */
+
+package ritabrata;
+
+public class DiffBetnMaxxOddMaxEven {
+	
+	
+	void getDiffBetnMaxOddMaxEven(int[] input) {
+		int even= input[0];
+		int odd= input[0];
+		for(int i=0; i<input.length; i++) {
+			if(input[i]%2==0) {
+				if(even<input[i])
+					even=input[i];
+			}
+				else {
+					if(odd<input[i])
+						odd=input[i];
+				}
+				
+		}
+		if (even > odd) {
+			int maxDif = even - odd;
+			System.out.println("Difference between max odd and max even number is :" + even + " - " + odd + " = "
+					+ maxDif);
+		} else {
+			int maxDif = odd - even;
+			System.out.println("Difference between max odd and max even number is :" + odd + " - " + even + " = "
+					+ maxDif);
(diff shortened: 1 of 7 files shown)


src/ritabrata/DiffBetnMaxxOddMaxEven.java
```diff
@@ -0,0 +1,49 @@
...
+ program 4 : print difference between max odd number and max even number from given array. difference should be positive
+
+int[] input = {1,11,44,23,55,99,23,22};
+hint : 99 - 44
+output : 55
+ */
+
+package ritabrata;
+
+public class DiffBetnMaxxOddMaxEven {
```
surabhi-9, line 11: will fail for given input:
99,11,1,23,55,44,23,22
to allow our pgm to run in all validation, please consider input[0] as either even or odd, if both is assigned to even and odd variables, pgm will fail for some inputs

src/ritabrata/FindMinimun.java
```diff
@@ -0,0 +1,33 @@
+/*
+  Assignment:14 15th Aug'2021
+ program 2 : find minimum number from given array.
+int[] input = {1,11,44,23,55,99,23};
+output : 11*/
+
+
```
surabhi-9, line 7: This logic will fail for below input:
{99,11,1,23,55,44,23,22}
please consider input[0] as either even or odd, you can also imagine one number from array can it be both?
so we have to assign input[0] either even or odd, other can make it to 0

src/ritabrata/FindMinimun.java
```diff
@@ -0,0 +1,33 @@
...
+package ritabrata;
+
+public class FindMinimun {
+	int findMin(int[] input) {
+		int number=0;
+		for(int i=0; i<input.length; i++) {
```
surabhi-9, line 13: code logic can be simplified and in less complexity as below:
int getMinValue(int[] number) {
		int min=number[0];
		for(int index=1;index<number.length;index++) {
			if(number[index]<min) {
				min=number[index];
			}
		}
		return min;
	}

src/ritabrata/MaxNumber.java
```diff
@@ -0,0 +1,31 @@
...
+program 1 : find maximum number from given array.
+
+int[] input = {1,11,44,23,55,99,23};
+output : 99
+*/
+
+
+package ritabrata;
+
+public class MaxNumber {
```
surabhi-9, line 11: same comment as min:
int getMaxValue(int[] number) {
		int max=number[0];
		for(int index=1;index<number.length;index++) {
			if(number[index]>max) {
				max=number[index];
			}
		}
		return max;
	}

surabhi-9 (approved):
approving for now. Please correct it as per comments

PR #771, opened 2021-08-16
Ritabrata assignment16

Ritabrata1996:
image


surabhi-9 (approved):
Please correct the codes again where same value is assigned to two variables e.g even = arr[0], odd=arr[0]

PR #772, opened 2021-08-16
Rasika_Assignment_11And12

RasikaNitin:
image
image

### DIFF
--- a/src/rasika/Assignment11.java
+++ b/src/rasika/Assignment11.java
@@ -1,0 +1,51 @@
+package rasika;
+
+class Assignment11{
+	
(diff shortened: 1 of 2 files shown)


src/rasika/Assignment11.java
```diff
@@ -0,0 +1,51 @@
+package rasika;
+
+class Assignment11{
+	
+	void printFirstNPrimeNum(int num){
```
surabhi-9, line 5: prime logic is wrong and giving wrong output, please correct and push this code again

surabhi-9 (approved):

PR #777, opened 2021-08-16
Shreya_Assignment_13

ShreyaGulhane:
image

### DIFF
--- a/src/shreya/ArrayProgram13.java
+++ b/src/shreya/ArrayProgram13.java
@@ -1,0 +1,24 @@
+// Program to return count of positive numbers from given array.
+package shreya;
+
+public class ArrayProgram13 {
+	static int count=0;
+	int getCountPositiveNumbers(int[] num) {
+		for(int index=0;index<num.length;index++) {
+			if(num[index]>0)
+				count++;
+		}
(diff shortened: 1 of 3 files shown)


src/shreya/ArrayProgram13.java
```diff
@@ -0,0 +1,24 @@
+// Program to return count of positive numbers from given array.
+package shreya;
+
+public class ArrayProgram13 {
+	static int count=0;
```
Upasana1703, line 5: No need to use instance variable and also if you are using the instance variable then it should not be static. As it depends on the input array and it may different for different objects.

src/shreya/ArrayProgram2_13.java
```diff
@@ -0,0 +1,24 @@
+// Program to count number of zero's from given array.
+package shreya;
+
+public class ArrayProgram2_13 {
+	static int count=0;
```
Upasana1703, line 5: No need to use instance variable and also if you are using the instance variable then it should not be static. As it depends on the input array and it may different for different objects.

Upasana1703 (approved):
Code looks good to me.

PR #778, opened 2021-08-16
Swati_Assignment_14

swatis9:
program 1 Output : find maximum number from given array.

image

program 2  Output : find minimum number from given array.

image

program 3 : Output : print difference between max and min number from given array.
image

program 4 Output  : print difference between max odd number and max even number from given array. difference should be positive
image


harshvegada (approved):
Code Logic looks good to me!

PR #801, opened 2021-08-16
TanmoyRoy_Assignment15

tanmoyroy96:
Output for wap that will return difference of even length name and odd lenght name.:
image

Output for program 2 : return middle char of all elements of the String array.
image

Output for program 3 : return difference between sum of even number - sum of odd numbers. Difference has to be positive. print the answer in main method.
image

### DIFF
--- a/src/tanmoy_Roy/Assignment15_1.java
+++ b/src/tanmoy_Roy/Assignment15_1.java
@@ -1,0 +1,37 @@
+package tanmoy_Roy;
+
+public class Assignment15_1 {
+	
+	int diffEvenAndOddLength(String[] array,int N) {
+		int count=0;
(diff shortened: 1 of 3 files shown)


src/tanmoy_Roy/Assignment15_1.java
```diff
@@ -0,0 +1,37 @@
+package tanmoy_Roy;
+
+public class Assignment15_1 {
+	
+	int diffEvenAndOddLength(String[] array,int N) {
```
harshvegada, line 5: same as don't require to pass length of array in method arg use it's property

harshvegada (approved):
Variable name should be small, best practices
Code Logic looks good to me!

PR #802, opened 2021-08-16
Sneha assignment 14

snehadanave:
image
image
image
image

### DIFF
--- a/src/sneha/Assignment_14_1.java
+++ b/src/sneha/Assignment_14_1.java
@@ -1,0 +1,28 @@
+// find maximum number from array
+
+package sneha;
+
+public class Assignment_14_1 {
+	
+	int getMaximum(int[] arr) {
+		int maximum = 0;
+		
+		for(int index = 0; index<arr.length; index++) {
+			if(arr[index] > maximum) {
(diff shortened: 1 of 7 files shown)


src/sneha/Assignment_14_1.java
```diff
@@ -0,0 +1,28 @@
+// find maximum number from array
+
+package sneha;
+
+public class Assignment_14_1 {
+	
+	int getMaximum(int[] arr) {
+		int maximum = 0;
```
Upasana1703, line 8: If all the elements of array are in negative then this logic will not work.

src/sneha/Assignment_14_3.java
```diff
@@ -0,0 +1,42 @@
+// print differnece between maximum and minimum from array
+
+package sneha;
+
+public class Assignment_14_3 {
+	
+	int getMaximum(int[] arr) {
+		int maximum = 0;
```
Upasana1703, line 8: If all the elements of array are less than zero than this logic will not work

Upasana1703 (approved):
Code looks good to me.

PR #805, opened 2021-08-16
paresh_assignment14

paresh129:
image

### DIFF
--- a/src/paresh/Assignment14.java
+++ b/src/paresh/Assignment14.java
@@ -1,0 +1,68 @@
+package paresh;
+
+public class Assignment14 {
+
+	int getMaxNumberFromArray(int[] num) {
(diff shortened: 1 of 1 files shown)


src/paresh/Assignment14.java
```diff
@@ -0,0 +1,68 @@
...
+				minimumNumber = arr[index];
+
+		}
+		return minimumNumber;
+	}
+
+	int getMaximumxAndMinimumNumberDiffrence(int[] num) {
+		int min = num[0];
+		int max = 0;
+		for (int index = 0; index < num.length; index++) {
```
SandeshMolane, line 27: we can use the functions which we wrote above to increase re-usability.

SandeshMolane (approved):
Code looks good to me!

PR #838, opened 2021-08-17
Apurva_Assignment_7_Loop

ApurvaD25:
WAP with For loop :- to find Even numbers within range, print all numbers which is divisible by 5, print all numbers which is divisible by 5 and divisible by 3, print all numbers which is divisible by 7 or 13. , sum of all the numbers in user define range. , difference between sum of odd numbers and even numbers in a given range 

image


harshvegada (approved):
Code Logic looks good to me!
Keep doing well

PR #843, opened 2021-08-17
Sayliv assignment 14

Saylivyavhare:
image

### DIFF
--- a/src/sayli_Vyavhare/Assignment14.java
+++ b/src/sayli_Vyavhare/Assignment14.java
@@ -1,0 +1,65 @@
+package sayli_Vyavhare;
+
+public class Assignment14 {
+	
+	int [] input= {1,11,44,23,55,99,23};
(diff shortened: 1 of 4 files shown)


src/sayli_Vyavhare/Assignment14.java
```diff
@@ -0,0 +1,65 @@
...
+	
+	void getDifferenceOfMaxOddAndMaxEven(int arr[]) {
+		int maxNumber1=0,maxEvenNumber=0,maxOddNumber=0,differenceEvenOdd=0;
+		
+		for(int index=0;index<arr.length;index++){
+				 if(maxNumber1<arr[index]) {
+				    maxNumber1=arr[index];
+				      if(maxNumber1% 2 ==0)
+				 		  maxEvenNumber=maxNumber1;
+				 	  if(maxNumber1% 2 !=0)
```
nehatawari, line 41: instead of two if condition u can use if else

nehatawari (approved):

PR #853, opened 2021-08-17
Sayaliv_Assignment_9

sayalivt:
image

### DIFF
--- a/src/sayali_V/Assignment_9.java
+++ b/src/sayali_V/Assignment_9.java
@@ -1,0 +1,78 @@
+package sayali_V;
+
+public class Assignment_9 
+{
+	static int iniMagQ, iniDosaQ, iniOilPouchesQ, iniPanipuriQ, iniMasalaQ;
+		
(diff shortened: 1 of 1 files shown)


src/sayali_V/Assignment_9.java
```diff
@@ -0,0 +1,78 @@
...
+	{
+		if (iniMagQ > 0 || iniDosaQ > 0 || iniOilPouchesQ > 0 || iniPanipuriQ > 0 || iniMasalaQ > 0)
+		{
+			iniMagQ = iniMagQ - purMagQ;
+			iniDosaQ = iniDosaQ - purDosaQ;
+			iniOilPouchesQ = iniOilPouchesQ - purOilPouchesQ;
+			iniPanipuriQ = iniPanipuriQ - purPanipuriQ;
+			iniMasalaQ = iniMasalaQ - purMasalaQ;
+		}
+			if (iniMagQ == 0)
```
nehatawari, line 26: for out of stock already we have method so no need to print here again

nehatawari:
try to cover negative scenarios also

PR #855, opened 2021-08-17
Sagar_Assignment14

sagaanna:
image

### DIFF
--- a/src/sagar_Anasane/Assignment14.java
+++ b/src/sagar_Anasane/Assignment14.java
@@ -1,0 +1,89 @@
+/* Assignment 14 : 15th Aug'2021
+program 1 : find maximum number from given array.
+int[] input = {1,11,44,23,55,99,23};
+output : 99
+program 2 : find minimum number from given array.
+int[] input = {1,11,44,23,55,99,23};
+output : 11
+program 3 : print difference between max and min number from given array.
+int[] input = {1,11,44,23,55,99,23};
+output : 98
+program 4 : print difference between max odd number and max even number from given array. difference should be positive
+int[] input = {1,11,44,23,55,99,23,22};
+hint : 99 - 44
+output : 55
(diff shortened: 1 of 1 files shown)


src/sagar_Anasane/Assignment14.java
```diff
@@ -0,0 +1,89 @@
...
+	
+	int printDiffBetweenMaxOddNosAndMaxEvenNos() {
+		int printDiff;
+		for (int index = 0;index<input1.length;index++) {
+			if(input1[index]%2==0 && input1[index] > maxEvenNum) {
+				maxEvenNum = input1[index];
+		    }
+		}
+	
+		for (int index = 0;index<input1.length;index++) {
```
nehatawari, line 61: no need of second loop

src/sagar_Anasane/Assignment14.java
```diff
@@ -0,0 +1,89 @@
...
+		System.out.println(maxNum-minNum);
+		return maxNum-minNum;
+	}
+	
+	int printDiffBetweenMaxOddNosAndMaxEvenNos() {
+		int printDiff;
+		for (int index = 0;index<input1.length;index++) {
+			if(input1[index]%2==0 && input1[index] > maxEvenNum) {
+				maxEvenNum = input1[index];
+		    }
```
nehatawari, line 58: int printDiffBetweenMaxOddNosAndMaxEvenNos() {
		int printDiff;
		for (int index = 0;index<input1.length;index++) {
			if(input1[index]%2==0 && input1[index] > maxEvenNum) {
				maxEvenNum = input1[index];
		    }else if(input1[index]%2!=0 && input1[index] > maxOddNum)
		    	maxOddNum = input1[index];
		}

instead of second loop it can be achieve in single loop

nehatawari (approved):

PR #865, opened 2021-08-17
Swati_Assignment_15

swatis9:
program 1 Output:  wap that will return difference of even length name and odd length name.

image

program 2 Output : return middle char of all elements of the String array.

image

program 3 Output: return difference between sum of even number - sum of odd numbers. Difference has to be positive. print the answer in main method.

image


harshvegada (approved):
Code Logic looks good to me!

PR #867, opened 2021-08-17
Sameer_Assignment_15

SZalke:


src/sameer/ArrayProgram10.java
```diff
@@ -0,0 +1,39 @@
...
+public class ArrayProgram10 {
+	int getDiffBetEvenSumAndOddSum(int[] arr) {
+		int difference = 0;
+		int evenSum = 0;
+		int oddSum = 0;
+		for (int index = 0; index < arr.length; index++) {
+			if (arr[index] % 2 == 0) {
+				evenSum = evenSum + arr[index];
+			}
+			if (arr[index] % 2 != 0) {
```
nehatawari, line 19: u can use if  else here instead of two if

nehatawari (approved):

PR #868, opened 2021-08-17
Apurva_Assignment_11_WhileLoop

ApurvaD25:
WAP using while loop: find first 10 prime numbers, count and add the number upto sum goes beyond limit 100

image


harshvegada (approved):
Code Logic looks good to me!

PR #872, opened 2021-08-17
Trupti assignment 10

truptimkulkarni:
image

### DIFF
--- a/src/trupti/Assignment_10.java
+++ b/src/trupti/Assignment_10.java
@@ -1,0 +1,53 @@
+package trupti;
+
+public class Assignment_10 {
+	int sum;
+	int primeCount;
+	int average;
+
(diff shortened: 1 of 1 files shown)


src/trupti/Assignment_10.java
```diff
@@ -0,0 +1,53 @@
...
+			if (num % index == 0) {
+				status = false;
+				break;
+			}
+		}
+		if (status) {
+			System.out.println(num + " is a prime number");
+			primeCount++;
+			sum += num;
+			average = sum / primeCount;
```
harshvegada, line 20: this not require, like for every prime u check avg
you can do this at end of iterations

harshvegada (approved):
`Code` Logic looks good to me!`

PR #908, opened 2021-08-17
Assignment13and14

LaxmanRDec19:
image

### DIFF
--- a/src/laxman/Assignment13.java
+++ b/src/laxman/Assignment13.java
@@ -1,0 +1,67 @@
+package laxman;
+
+public class Assignment13 {
+		
+	int getPositiveNumber(int[]input){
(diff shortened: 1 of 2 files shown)


src/laxman/Assignment13.java
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+		
+	
+
+		public static void main(String[]args) {
+			Assignment13 assignment13 =new Assignment13();
+			int [] arr = new int[8];
+			int [] num = new int[8];
+			arr[0]=1;
```
harshvegada, line 43: instead of this, you can define this like
int a[] = {elements}

harshvegada (approved):
Code logic looks good to me!

PR #911, opened 2021-08-18
SayliVyavhare_Assignment_16

Saylivyavhare:
image

### DIFF
--- a/src/sayli_Vyavhare/Assignment_16.java
+++ b/src/sayli_Vyavhare/Assignment_16.java
@@ -1,0 +1,24 @@
+package sayli_Vyavhare;
+
+public class Assignment_16 {
+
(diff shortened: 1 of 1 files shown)


src/sayli_Vyavhare/Assignment_16.java
```diff
@@ -0,0 +1,24 @@
...
+				visibleTree=treeHeight[index];
+			    count++;			
+			}
+		}
+		return count;		
+	}
+	
+	public static void main(String[] args) {
+		Assignment_16 assignment_16=new Assignment_16();
+		int visibleCount=assignment_16.getCountOfVisibleTree();
```
nehatawari, line 21: Visible count is declared for two time at line no 5 and 21.it should be declared once

nehatawari (approved):

PR #927, opened 2021-08-18
Riten assignment 16

ritenthakkar:
image

### DIFF
--- a/src/riten/VisibleTree.java
+++ b/src/riten/VisibleTree.java
@@ -1,0 +1,31 @@
+package riten;
+
+class VisibleTree {
+
+	int count = 0;
+
+	void findCount(int[] arr) {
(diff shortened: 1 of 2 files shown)


src/riten/VisibleTree.java
```diff
@@ -0,0 +1,31 @@
...
+		int visiblecount = 0;
+		for (int index = 0; index < arr.length; index++) {
+			if (arr[index] > visiblecount) {
+				visiblecount = arr[index];
+				count++;
+			} else if (arr[index] == visiblecount) {
+				visiblecount = arr[index];
+			}
+
+			else if (arr[index] < visiblecount) {
```
surabhi-9, line 17: not needed to write this block. rest code looks better

surabhi-9 (approved):

PR #943, opened 2021-08-18
Sneha_Assignment_17

snehadanave:
image

### DIFF
--- a/src/sneha/Assignment_17.java
+++ b/src/sneha/Assignment_17.java
@@ -1,0 +1,32 @@
+/* Write a method which takes one String parameter and return a reverse String.
+
+Hint : String getReverseString(String input){
+
+return output;
+}
+
+input : technocredits
+output : stiderconhcet */
+
+package sneha;
+
+public class Assignment_17 {
+
+	String getReverseString(String input){
+	char[] string = input.toCharArray();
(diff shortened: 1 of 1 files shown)


src/sneha/Assignment_17.java
```diff
@@ -0,0 +1,32 @@
...
+			System.out.print(string[i]);
+	}
+	String output = null;
+	
+		return output;
+	}
+	
+	public static void main(String[] a) {
+		Assignment_17 assignment_17 = new Assignment_17();
+		String output = null;
```
Upasana1703, line 29: No need of line 29

src/sneha/Assignment_17.java
```diff
@@ -0,0 +1,32 @@
...
+public class Assignment_17 {
+
+	String getReverseString(String input){
+	char[] string = input.toCharArray();
+	System.out.println("Your given string is : " + input);
+	System.out.println("String after reverse is:");
+	for(int i = string.length-1; i >= 0 ; i--) {
+			System.out.print(string[i]);
+	}
+	String output = null;
```
Upasana1703, line 22: Whats the use of declaring output as null and then returning it.

No need of line 22 and 23

harshvegada (approved):
Code Logic looks good to me!

PR #950, opened 2021-08-18
Suchita_Assignment_17

shuchkrish:
image

### DIFF
--- a/src/suchita/Assignment17.java
+++ b/src/suchita/Assignment17.java
@@ -1,0 +1,20 @@
+package suchita;
+
+class Assignment17 {
+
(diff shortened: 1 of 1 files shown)


src/suchita/Assignment17.java
```diff
@@ -0,0 +1,20 @@
+package suchita;
+
+class Assignment17 {
+
+	String getReverseString(String arr) {
+		
+		String reverse = " ";
```
Upasana1703, line 7: As you are concatenating the characters on space value. so in the reverse string, first character will be space.

Upasana1703 (approved):
Code looks good to me

PR #970, opened 2021-08-18
Shalaka_reverseString_Assignment17

Shalakaprayag15:
image

### DIFF
--- a/src/shalaka/ArrayPrograms/ReverseString_Assignment17.java
+++ b/src/shalaka/ArrayPrograms/ReverseString_Assignment17.java
@@ -1,0 +1,30 @@
+
+/*Write a method which takes one String parameter and return a reverse String.
+
+
+Hint : String getReverseString(String input){
+
+return output;
+}
+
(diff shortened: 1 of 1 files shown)


src/shalaka/ArrayPrograms/ReverseString_Assignment17.java
```diff
@@ -0,0 +1,30 @@
...
+			output = output+input.charAt(index);
+		}
+		return output;
+	}
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+		String input = "technocredits";
+		ReverseString_Assignment17 reverseString_Assignment17 = new ReverseString_Assignment17();
+		String str = reverseString_Assignment17.getReverseString(input);
```
nehatawari, line 27: try not to use variable until u are not performing any action on that. str is not needed you can directly print reverseString_Assignment17.getReverseString(input)

nehatawari (approved):

PR #993, opened 2021-08-18
Shyam_Assignment_17

shyamanasane:
image

### DIFF
--- a/src/shyam/ArrayExample2.java
+++ b/src/shyam/ArrayExample2.java
@@ -1,0 +1,27 @@
+package shyam;
+
+public class ArrayExample2{
+	
+	int countOfZeros;
+
+	public static void main(String[] args) {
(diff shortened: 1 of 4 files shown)


src/shyam/ArrayExample2.java
```diff
@@ -0,0 +1,27 @@
+package shyam;
+
+public class ArrayExample2{
+	
+	int countOfZeros;
```
Upasana1703, line 5: Try to do this without creating instance variable.

src/shyam/ArrayExample3.java
```diff
@@ -0,0 +1,20 @@
+package shyam;
+
+public class ArrayExample3 {
+
+	void createAndPrintArray(int input) {
```
Upasana1703, line 5: According to question, you should have returned array from this method.

PR #997, opened 2021-08-18
Monali_Assignment_17

Monaliborole:
image

### DIFF
--- a/src/monali/ReverseString.java
+++ b/src/monali/ReverseString.java
@@ -1,0 +1,28 @@
+package monali;
+
+public class ReverseString {
+
+	String getReverseString(String str) {
+
+		String reverseStr = "";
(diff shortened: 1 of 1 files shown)


src/monali/ReverseString.java
```diff
@@ -0,0 +1,28 @@
...
+		}
+		return reverseStr;
+	}
+
+	String processData() {
+		String str = "Technocredits";
+		return getReverseString(str);
+	}
+
+	void displayReverseString() {
```
harshvegada, line 19: you can direct call the getReverseString method from main

harshvegada (approved):
Code Logic looks good to me!

src/monali/ReverseString.java
Monaliborole: yes but for more practice purpose i returned it another method and capture it in main.

PR #1029, opened 2021-08-19
Dattaraj Assignment 16

dattaraj123456:
image

### DIFF
--- a/src/dattaraj/Assignment_16.java
+++ b/src/dattaraj/Assignment_16.java
@@ -1,0 +1,22 @@
+package dattaraj;
+
+public class Assignment_16 {
+	int geVisibleTreeCnt(int[] input) {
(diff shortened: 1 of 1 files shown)


src/dattaraj/Assignment_16.java
```diff
@@ -0,0 +1,22 @@
+package dattaraj;
+
+public class Assignment_16 {
+	int geVisibleTreeCnt(int[] input) {
+		int counter = 1;
+		int maxHeight = input[0];
+		for (int index = 0; index < input.length; index++) {
```
harshvegada, line 7: as you already taken maxHeight as input[0] then u ideally should start loop from 1

harshvegada (approved):
Code Logic looks good to me!

PR #1040, opened 2021-08-19
Chandni assignment 16

ChandniBhojwani:
image

### DIFF
--- a/src/chandni_bhoj/Assignment_13.java
+++ b/src/chandni_bhoj/Assignment_13.java
@@ -1,0 +1,67 @@
+package chandni_bhoj;
+
+public class Assignment_13 {
+
+	int getCountOfPositiveNumber (int [] arr) {
+		int count = 0;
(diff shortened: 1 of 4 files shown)


harshvegada:
Screenshot 2021-08-19 at 10 44 06 PM
As I drop message on skype do not create a branch from another branch

src/chandni_bhoj/Assignment_13.java
```diff
@@ -0,0 +1,67 @@
...
+				System.out.print (index + " , ");
+			index--;
+
+		}
+	}
+
+	public static void main(String[] args) {
+		Assignment_13 assignment_13 = new Assignment_13 ();
+		int [] input = new int [8];
+		input[0] = 1;
```
harshvegada, line 39: you should create array like this
int a[] ={3,4,5,6,7} instead of a[0] = 8 etc..

harshvegada (approved):
Code Logic looks good to me!

PR #1077, opened 2021-08-19
akankshak_Assignment_17

akankshaz:
image

### DIFF
--- a/src/akanksha_K/Assignment17.java
+++ b/src/akanksha_K/Assignment17.java
@@ -1,0 +1,18 @@
+package akanksha_K;
+
+public class Assignment17 {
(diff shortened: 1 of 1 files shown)


src/akanksha_K/Assignment17.java
```diff
@@ -0,0 +1,18 @@
+package akanksha_K;
+
+public class Assignment17 {
+	
+	String getreverseString(String input) {
+		String output = " ";
```
harshvegada, line 6: you should remove this blank space in output
String output = "";

harshvegada (approved):
Code logic looks good to me!

PR #1083, opened 2021-08-19
ankita_assignment_15

Avkene:
image
image
image

### DIFF
--- a/src/ankita/Evenoddlength.java
+++ b/src/ankita/Evenoddlength.java
@@ -1,0 +1,22 @@
+package ankita;
+
+public class Evenoddlength {
+	int diffLength (String [] arr) 
+	{
+		int evensum=0;
+		int oddsum=0;
+		for(int index=0;index<arr.length;index++) 
+		if(arr[index].length()%2==0)
+			evensum+=arr[index].length();
+			else
+			oddsum+=arr[index].length();
+		return evensum-oddsum;
+	}
+	
+	public static void main(String[] args) {
+		Evenoddlength evenoddlength=new Evenoddlength();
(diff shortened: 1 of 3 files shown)


src/ankita/Evenoddlength.java
```diff
@@ -0,0 +1,22 @@
...
+	int diffLength (String [] arr) 
+	{
+		int evensum=0;
+		int oddsum=0;
+		for(int index=0;index<arr.length;index++) 
+		if(arr[index].length()%2==0)
+			evensum+=arr[index].length();
+			else
+			oddsum+=arr[index].length();
+		return evensum-oddsum;
```
harshvegada, line 13: You should check whether oddSum is higher or evenSum based on that you should write condition so that returning value will be positive
if(oddSum > evenSum)
	return (oddSum - evenSum)
else
	return (evenSum - oddSum)

src/ankita/Middlechar.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class Middlechar {
+	int middchar;
+	void getmiddChar(String[] arr) {
+		char [] ch=new char[3];
+	 for(int index=0;index <arr.length;index++)
+		 if(arr[index].length()%2==0) {
+			 middchar=arr[index].length()/2-1;
+			 ch[index]=arr[index].charAt(middchar);
+			 System.out.println(ch);
```
harshvegada, line 11: > Here u are printing the ch(which is array, n if print array then it will memory address not the value).
> you should create temp `char ch` that ch variable should have that single character value

harshvegada (approved):

PR #1087, opened 2021-08-19
Indrani_Assignment_18

Indrani0201:
Program 1
image
Program 2
image

### DIFF
--- a/src/indrani/Assignment_18_program2.java
+++ b/src/indrani/Assignment_18_program2.java
@@ -1,0 +1,33 @@
+package indrani;
+
+import java.util.Arrays;
+
+     public class Assignment_18_program2{
+
+      String getReverseString(String input){
+	     String output="";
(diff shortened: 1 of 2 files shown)


src/indrani/Assignment_18_program2.java
```diff
@@ -0,0 +1,33 @@
...
+	        }
+	        return output;
+            }
+
+       String[] getReverseArrayReverseOrder(String[] input){
+	   String[] output=new String[input.length];
+	   int n=input.length-1;
+	   for(int index=0;index<=input.length-1;index++){
+		     output[index]=getReverseString(input[n]);
+		      n--;
```
harshvegada, line 20: Here you can write like below, n you don't require to write n++ explicitly
`output[index]=getReverseString(input[n--]);`

harshvegada (approved):

PR #1089, opened 2021-08-19
Monali_Assignment_18

Monaliborole:
image

image

### DIFF
--- a/src/monali/ReverseOrderStringReverse.java
+++ b/src/monali/ReverseOrderStringReverse.java
@@ -19,10 +19,10 @@ public class ReverseOrderStringReverse {
 	}
 
 	String[] getStringReverseArray(String[] input) {
-	
-		int resultIndex=0;
+
+		int resultIndex = 0;
 		String output[] = new String[input.length];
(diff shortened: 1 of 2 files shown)


src/monali/ReverseOrderStringReverse.java
```diff
@@ -0,0 +1,38 @@
...
+		}
+		return output;
+	}
+
+	String[] getStringReverseArray(String[] input) {
+
+		int resultIndex = 0;
+		String output[] = new String[input.length];
+		for (int index = input.length - 1; index >= 0; index--) {
+			output[resultIndex] = getReverseString(input[index]);
```
harshvegada, line 26: Here you can write like below, n you don't require to write resultIndex++ explicitly
`output[resultIndex ++]=ietReverseString(input[index]);`

harshvegada (approved):

src/monali/ReverseOrderStringReverse.java
Monaliborole: Thanks Harsh i will definitely improve it in next assignments.

PR #1098, opened 2021-08-20
Arti_G_Assignment_18

Arti-Saini:
image
image

### DIFF
--- a/src/arti_G_Assignment18/ReverseOderString.java
+++ b/src/arti_G_Assignment18/ReverseOderString.java
@@ -1,0 +1,40 @@
+package arti_G_Assignment18;
+
+/*Assignment - 18 : 19th Aug'21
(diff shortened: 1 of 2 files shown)


src/arti_G_Assignment18/ReverseOderString.java
```diff
@@ -0,0 +1,40 @@
...
+		return output;
+	}
+
+	// 2 program
+	String[] reverseOderPlaceInReverseOder(String[] input) {
+		String[] output = new String[input.length];
+		int num = input.length-1;
+		for (int index = 0; index < input.length; index++) {
+			output[index] = reverseString(input[num]);
+			num--;
```
harshvegada, line 28: Here you can write like below, n you don't require to write index-- explicitly
`output[index]=input[num--];`

harshvegada (approved):

PR #1100, opened 2021-08-20
Tarang branch

tarang1162:


src/tarang/Assignment_9.java
```diff
@@ -0,0 +1,146 @@
...
+	void getOrderDetails(int maggiePackate, int setDosa, int pouchesP, int panipuriP, int masalaP) {
+
+		if (maggie > 0) {
+			if (maggiePackate <= maggie) {
+				maggie = maggie - maggiePackate;
+
+			} else {
+				System.out.println("Dont have sufficient quantity of maggie.Requested maggie is " + maggiePackate
+						+ " but available qauntity is " + maggie);
+				maggiePackate = 0;
```
harshvegada, line 28: maggiePackate = 0;

if printed message correct but making this variable 0, not getting

like agr aap store jate n u ask for 51 pack but shop owner have 50 only then owner not going make it 0 right?
If you trying to say that owner will provide u 50 n then making stock as 0 then it make sense and for that u need to update your message as well..

this will applicable to other conditions as well.

harshvegada (approved):
Cod looks good to me!

PR #1101, opened 2021-08-20
Suruchi_Assignment_19

suruGIT21:
**Program 1:**
image

**Program 2:**
image

**Program 3:**
image

**Program 4:**
image

### DIFF
--- a/src/suruchi/Assignment19_Program1.java
+++ b/src/suruchi/Assignment19_Program1.java
@@ -1,0 +1,21 @@
+package suruchi;
+
+public class Assignment19_Program1 {
+
+	int getCountOfWordsInString(String str) {
+		String[] arr = str.split(" ");
+		int count = 0;
+		for (int index = 0; index < arr.length; index++)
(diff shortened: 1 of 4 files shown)


src/suruchi/Assignment19_Program1.java
```diff
@@ -0,0 +1,21 @@
+package suruchi;
+
+public class Assignment19_Program1 {
+
+	int getCountOfWordsInString(String str) {
+		String[] arr = str.split(" ");
+		int count = 0;
+		for (int index = 0; index < arr.length; index++)
```
harshvegada, line 8: Instead of writing a loop, you should use a split(“ “) function which gives an array, and on array, u can use the length property. you don't require for loop

harshvegada (approved):

src/suruchi/Assignment19_Program1.java
suruGIT21: Okay got it.

PR #1104, opened 2021-08-20
AkanshaJain_Assignment_19 and refactoring

akansha-jain6:
Remove one line from my existing Assignment 15 Program 3 which is added mistakenly.

Assignment_19
Screenshot 2021-08-20 at 3 00 00 PM

### DIFF
--- a/src/akanksha_Jain/Assignment_19.java
+++ b/src/akanksha_Jain/Assignment_19.java
@@ -3,13 +3,8 @@ package akanksha_Jain;
 public class Assignment_19 {
 
 	void getWordsFromString(String input) {
-		int count = 0;
-		for(int index=0; index<input.length()-1; index++) {
-			char ch = input.charAt(index);
-			if (ch == ' ')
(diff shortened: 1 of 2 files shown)


src/akanksha_Jain/Assignment_19.java
```diff
@@ -0,0 +1,60 @@
+package akanksha_Jain;
+
+public class Assignment_19 {
+
+	void getWordsFromString(String input) {
+		int count = 0;
+		for(int index=0; index<input.length()-1; index++) {
```
harshvegada, line 7: if u use split(" ") function then u don't require for loop

harshvegada (approved):

src/akanksha_Jain/Assignment_19.java
akansha-jain6: > if u use split(" ") function then u don't require for loop

@harshvegada I have changed this logic. Thank you!

PR #1106, opened 2021-08-20
Elizabeth_Assignment_19

Elizmary:
image

### DIFF
--- a/src/elizabeth/Assignment19.java
+++ b/src/elizabeth/Assignment19.java
@@ -1,0 +1,99 @@
+/*program 1: Find how many words in a given statement.
+input = "Hi Techno Credits Hello"
+output : 4
+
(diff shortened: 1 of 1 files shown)


src/elizabeth/Assignment19.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Scanner;
+
+class Assignment19{
+	
+	int getWordCount(String str){
+		
+		String[] input= new String[str.length()];
+		input=str.split(" ");
+		int output=input.length;
+		return output;
```
harshvegada, line 26: you can write below
`String in[] = str.split(" ")`
`in.length`
 or
`str.split(" ").length`

harshvegada (approved):

PR #1109, opened 2021-08-20
Nishi_Assignment_19

NishiBhartii:
image
image
image
image

### DIFF
--- a/src/nishi/Assignment19B.java
+++ b/src/nishi/Assignment19B.java
@@ -1,0 +1,31 @@
+package nishi;
+
+import java.util.Scanner;
+
+public class Assignment19B {
+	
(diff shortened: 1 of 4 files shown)


src/nishi/Assignment19B.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Scanner;
+
+public class Assignment19B {
+	
+	String findMaximumWord(String input)
+	{
+		String[] output = input.split(" ");
+		String max=output[0];
+		for(int i=0;i<output.length;i++)
```
harshvegada, line 11: As u taken 
`String max=output[0];` then u should start the loop from 1

harshvegada (approved):

PR #1112, opened 2021-08-20
Madhavi_Assignment_18

Madhavi14C:
image
image

### DIFF
--- a/src/madhavi/Assignment_18_1.java
+++ b/src/madhavi/Assignment_18_1.java
@@ -1,0 +1,42 @@
+package madhavi;
+
+import java.util.Arrays;
+
+public class Assignment_18_1 {
+	String getReverseString(String input) {
+		String output = "";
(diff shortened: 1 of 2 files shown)


src/madhavi/Assignment_18_1.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	String[] getReverseStringArayAtReversePlace(String[] input) {
+
+		String[] output = new String[input.length];
+		int temp = 0;
+
+		for (int index = input.length - 1; index >= 0; index--) {
+			output[temp] = getReverseString(input[index]);
+			temp++;
```
harshvegada, line 23: Here you can write like below, n you don't require to write temp++ explicitly
`output[temp++]=getReverseString(input[index]);`

harshvegada (approved):

PR #1115, opened 2021-08-20
Himanshu_Assignment_19

himanshugpt56:
image
image
image
image

### DIFF
--- a/src/himanshu/Assignment_19_Program1.java
+++ b/src/himanshu/Assignment_19_Program1.java
@@ -1,0 +1,28 @@
+package himanshu;
+
+public class Assignment_19_Program1 {
+	
+	void usingSpiltMethod(String input){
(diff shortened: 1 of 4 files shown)


src/himanshu/Assignment_19_Program1.java
```diff
@@ -0,0 +1,28 @@
+package himanshu;
+
+public class Assignment_19_Program1 {
+	
+	void usingSpiltMethod(String input){
+		int count=0;
+		String[] arr = input.split(" ");
+		for(int index=0;index<arr.length;index++){
```
harshvegada, line 8: Instead of writing a loop, you should use a split(“ “) function which gives an array, and on array, u can use the length property

input.split(" ").length

harshvegada (approved):

PR #1123, opened 2021-08-20
Purshottam_Assignment19

purshottam94:
1)
image
2)
image
3)
image
4)
image

### DIFF
--- a/src/purshottamJoshi/Assignment19_1.java
+++ b/src/purshottamJoshi/Assignment19_1.java
@@ -1,0 +1,24 @@
+package purshottamJoshi;
+
+class Assignment19_1{
+	int count;
+	int countWords(String arr){
+		
+		for(int index=0; index <arr.length();index++){
+			char ch = arr.charAt(index);
+				if( ch == ' '){
+					count++;
+				}
+		}
+		return count+1;
+	}
+	
+	public static void main(String[] args){
+		String str = "Hi Techno Credits Hello";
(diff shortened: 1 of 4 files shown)


src/purshottamJoshi/Assignment19_1.java
```diff
@@ -0,0 +1,24 @@
+package purshottamJoshi;
+
+class Assignment19_1{
+	int count;
+	int countWords(String arr){
+		
+		for(int index=0; index <arr.length();index++){
```
harshvegada, line 7: Instead of writing a loop, you should use a split(“ “) function which gives an array, and on array, u can use the length property

src/purshottamJoshi/Assignment19_2.java
```diff
@@ -0,0 +1,22 @@
...
+		//int max =0;
+		String maxWord = "";
+		//System.out.println(str);
+		//Assignment19_2 assignment19_2 = new Assignment19_2();
+		String[] output = input.split(" ");
+			for(int index = 0 ;index < output.length;index++){
+				//if(arr[index])
+				if(output[index].length() > maxWord.length()){
+					maxWord=output[index];
+				System.out.println(output[index]+"-->"+output[index].length());
```
harshvegada, line 15: you should write SOP outside the loop, else it will print multiple times whenever condition true.

harshvegada (approved):

PR #1126, opened 2021-08-20
Madhavi_Assignment_19

Madhavi14C:
image
image
image
image

### DIFF
--- a/src/madhavi/Assignment_19_1.java
+++ b/src/madhavi/Assignment_19_1.java
@@ -1,0 +1,23 @@
+package madhavi;
+
+public class Assignment_19_1 {
+	int getStringWordCount(String input) {
+		int count = 0;
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
+			if (ch == ' ') {
+				count++;
+			}
+		}
+
+		return ++count;
+	}
+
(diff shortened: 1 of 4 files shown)


src/madhavi/Assignment_19_1.java
```diff
@@ -0,0 +1,23 @@
+package madhavi;
+
+public class Assignment_19_1 {
+	int getStringWordCount(String input) {
+		int count = 0;
+		for (int index = 0; index < input.length(); index++) {
```
harshvegada, line 6: Instead of writing a loop, you should use a split(“ “) function which gives an array, and on array, u can use the length property

src/madhavi/Assignment_19_4.java
```diff
@@ -0,0 +1,21 @@
+package madhavi;
+
+public class Assignment_19_4 {
+	int getSum(String input) {
+		String[] inpArray = input.split(" ");
+		int sum = 0;
+		for (int index = 0; index < inpArray.length; index++) {
+			int num1 = Integer.parseInt(inpArray[index]);
+			sum = sum + num1;
```
harshvegada, line 9: this can be written as below FYI,

you are string in variable n that can be direct add it to sum

harshvegada (approved):

PR #1134, opened 2021-08-20
Ritabrata_Assignment19

Ritabrata1996:
image

### DIFF
--- a/src/ritabrata/Assignment19.java
+++ b/src/ritabrata/Assignment19.java
@@ -1,0 +1,84 @@
+/*Assignment - 19 : 20th Aug'2021
+
+program 1: Find how many words in a given statement.
+input = "Hi Techno Credits Hello"
+output : 4
+
+program 2: Find maximum word length from given statement.
+input = "good morning technocredits hi hello";
+output : technocredits
+
+program 3 : Find minimum word length from given statement.
+input = "good morning technocredits hi hello";
+output : hi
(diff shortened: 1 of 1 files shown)


src/ritabrata/Assignment19.java
```diff
@@ -0,0 +1,84 @@
...
+input = "10 10 20 30";
+output : 70       */
+
+
+package ritabrata;
+
+public class Assignment19 {
+	int wordCount(String input) {
+		int count=0;
+		for(int index=0; index<input.length(); index++) {
```
harshvegada, line 25: Instead of writing a loop, you should use a split(“ “) function which gives an array, and on the array, u can use the length property.

src/ritabrata/Assignment19.java
```diff
@@ -0,0 +1,84 @@
...
+		System.out.println("The word with minimum no of letters is: "+minStr);
+		return minStr;
+	}
+	
+	int sumNumbers(String input) {
+		String[] arr= input.split(" ");
+		int sum=0;
+		for(int index=0; index<arr.length; index++) {
+			int num= Integer.parseInt(arr[index]);
+			sum=sum+num;
```
harshvegada, line 69: Just FYI,
sum=sum+Integer.parseInt(arr[index]);

harshvegada (approved):

PR #1135, opened 2021-08-20
Sourabh_Assignment_19

sourabhskulkarni:
image

### DIFF
--- a/src/sourabh/Assignment_19.java
+++ b/src/sourabh/Assignment_19.java
@@ -1,0 +1,66 @@
+package sourabh;
+
+public class Assignment_19{
+
+	//program 1: Find how many words in a given statement.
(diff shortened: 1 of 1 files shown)


src/sourabh/Assignment_19.java
```diff
@@ -0,0 +1,66 @@
...
+	 }
+
+	 //program 4: sum of all numbers in given statement.
+	 void getSumINtArray() {
+		 String srt ="10 25 38 73";
+		 String[] num= srt.split(" ");
+		 int sum=0;
+		 for (int index=0; index< num.length; index++) {
+			 int number= Integer.parseInt(num[index]);
+			 sum= number +number;
```
harshvegada, line 50: this can be written as below FYI,

harshvegada (approved):

PR #1140, opened 2021-08-20
sushant_Assignment19

sushant3689:
image

### DIFF
--- a/src/sushant/Assignment_19.java
+++ b/src/sushant/Assignment_19.java
@@ -1,0 +1,69 @@
+/*Assignment - 19 : 20th Aug'2021
+program 1: Find how many words in a given statement.
+input = "Hi Techno Credits Hello"
+output : 4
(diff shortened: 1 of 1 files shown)


src/sushant/Assignment_19.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	void sumOfAllNumbersInString(String inputString) {
+		System.out.println("---------------------------------------------");
+		System.out.println("Input String of numbers is :- "+ inputString);
+		String[] stringArray = inputString.split(" ");
+		int sum = 0;
+		for(int index = 0; index < stringArray.length; index ++) {
+			int number = Integer.parseInt(stringArray[index]);
+			sum += number;
```
harshvegada, line 63: this can be written as below FYI,

harshvegada (approved):

PR #1142, opened 2021-08-20
Chandani Assignment 19

chandni0604:
Program 1 : 
image

Program 2 : 
image

Program 3 : 
image

Program 4 : 
image

### DIFF
--- a/src/chandani/Assignment19a.java
+++ b/src/chandani/Assignment19a.java
@@ -1,0 +1,17 @@
+class Assignment19a{
+
+	int getCountOfWordsinString(String str){
+		int count = 0;
+		String[] arr = str.split(" ");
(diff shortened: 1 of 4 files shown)


src/chandani/Assignment19a.java
```diff
@@ -0,0 +1,19 @@
+package chandani;
+
+class Assignment19a{
+
+	int getCountOfWordsinString(String str){
+		int count = 0;
+		String[] arr = str.split(" ");
```
harshvegada, line 7: You can use are.lenth directly,not required for loop

harshvegada (approved):
Code looks good

src/chandani/Assignment19a.java
chandni0604: cool will take care about this :)

PR #1143, opened 2021-08-20
Swati_Assignment_19

swatis9:
program 1: Find how many words in a given statement.

image

program 2: Find maximum word length from given statement.

image

program 3 : Find minimum word length from given statement.

image

program 4: sum of all numbers in given statement.
image


harshvegada (approved):

PR #1150, opened 2021-08-20
akankshak_Assignment_19

akankshaz:
image

### DIFF
--- a/src/akanksha_K/Assignment19.java
+++ b/src/akanksha_K/Assignment19.java
@@ -1,0 +1,60 @@
+package akanksha_K;
+
+public class Assignment19 {
+	
+	static String[] findWords(String str) {
+		return str.split(" ");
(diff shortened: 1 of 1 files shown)


src/akanksha_K/Assignment19.java
```diff
@@ -0,0 +1,60 @@
...
+		return min_length_word;
+	}
+	
+	static int sum(String str) {
+		String words[] = findWords(str);
+		int sum = 0;
+		for (int index=0; index<words.length; index++) {
+			String word = words[index];
+			int val = Integer.parseInt(word);
+			sum = sum + val;
```
harshvegada, line 43: this can be written as below FYI,

sum = sum + Integer.parseInt(words[index]);

harshvegada (approved):

PR #1151, opened 2021-08-20
Margi_Assignment_16

MargiSantoki:
image

### DIFF
--- a/src/margi/Assignment_16.java
+++ b/src/margi/Assignment_16.java
@@ -1,0 +1,27 @@
+package margi;
+
+public class Assignment_16 {
+
+	//count visible trees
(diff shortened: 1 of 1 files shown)


src/margi/Assignment_16.java
```diff
@@ -0,0 +1,27 @@
+package margi;
+
+public class Assignment_16 {
+
+	//count visible trees
+	int cntVisibleTrees(int[] trees) {
+		int count=1;
+		int visible = trees[0];
+		for(int index=0; index < trees.length; index++) {
```
harshvegada, line 9: you can start index from 1 as first tree always visible to u.

harshvegada (approved):

PR #1152, opened 2021-08-20
SayliVyavhare_Assignment_18

Saylivyavhare:
image


src/sayli_Vyavhare/Assignment_18.java
```diff
@@ -0,0 +1,38 @@
+package sayli_Vyavhare;
+
```
nehatawari, line 2: Write problem statement for future reference

PR #1159, opened 2021-08-20
Avinash_Assignment_19

AvinashSPatil:
image


src/avinash/Assignmnet19.java
```diff
@@ -0,0 +1,115 @@
...
+			
+		}
+		return output;
+	}
+	
+	String minWordLength(String str2){
+		String output="";
+		String[] input= new String[str2.length()];
+		input=str2.split(" ");
+		int min=input[0].length();
```
harshvegada, line 53: As u taken 
`int min=input[0];` then u should start the loop from 1

harshvegada (approved):

PR #1160, opened 2021-08-20
TanmoyRoy_Assignment19

tanmoyroy96:
Output program 1: Find how many words in a given statement.
image

Output program 2 Find maximum word length from given statement.
image

Output program 3 Find minimum word length from given statement.
image

Output sum of all numbers in given statement.
image


harshvegada (approved):

PR #1168, opened 2021-08-20
Neha_Assignment_19

NehaJTechnocredits:
Program 1 Output:
image

Program 2 Output:
image

Program 3 Output:
image

Program 4 Output:
image

### DIFF
--- a/src/neha/Assignment19pg3.java
+++ b/src/neha/Assignment19pg3.java
@@ -1,0 +1,22 @@
+package neha;
+
+/*program 3 : Find minimum word length from given statement.
+input = "good morning technocredits hi hello";
+output : hi*/
+public class Assignment19pg3 {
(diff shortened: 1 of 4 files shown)


src/neha/Assignment19pg3.java
```diff
@@ -0,0 +1,22 @@
+package neha;
+
+/*program 3 : Find minimum word length from given statement.
+input = "good morning technocredits hi hello";
+output : hi*/
+public class Assignment19pg3 {
+	String minLengthWord(String strArray) {
+		String[] splitString = strArray.split(" ");
+		String minWord = splitString[0];
+		for (int index = 0; index < splitString.length; index++) {
```
harshvegada, line 10: As you take smallestString as splitString[0] then the loop can start from 1

harshvegada (approved):

PR #1174, opened 2021-08-21
Dattaraj Assignment 19

dattaraj123456:
Program 1
image

Program 2
image

Program 3
image

Program 4
image

### DIFF
--- a/src/dattaraj/Assignment_19_P3.java
+++ b/src/dattaraj/Assignment_19_P3.java
@@ -1,0 +1,20 @@
+package dattaraj;
+
+public class Assignment_19_P3 {
+	String minLengthWord(String strArray) {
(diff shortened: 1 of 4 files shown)


src/dattaraj/Assignment_19_P3.java
```diff
@@ -0,0 +1,20 @@
+package dattaraj;
+
+public class Assignment_19_P3 {
+	String minLengthWord(String strArray) {
+		String[] splitString = strArray.split(" ");
+		String minWord = splitString[0];
+		for (int index = 0; index < splitString.length; index++) {
```
harshvegada, line 7: As you take minWord as splitString[0] then loop can start from 1

harshvegada (approved):

PR #1183, opened 2021-08-21
AmolP_Assignment_20

amolpawar4334:
Program 1:

image

Program 2 :

image

Program 3:

image

Program 4:
image

### DIFF
--- a/src/amol_Pawar/Assignment_20/Program_4.java
+++ b/src/amol_Pawar/Assignment_20/Program_4.java
@@ -1,0 +1,29 @@
+package amol_Pawar.Assignment_20;
+
+public class Program_4 {
+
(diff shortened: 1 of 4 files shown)


src/amol_Pawar/Assignment_20/Program_4.java
```diff
@@ -0,0 +1,29 @@
...
+
+	void getDuplicateCharFromString(String input, char ch) {
+		int count = 0;
+		for (int index = 0; index < input.length(); index++) {
+
+			if (ch == input.charAt(index)) {
+				count++;
+			}
+		}
+		if (count > 1)
```
harshvegada, line 13: I don't think so this is require!

harshvegada (approved):

PR #1191, opened 2021-08-21
ShwetaD_Assignment_19

ShwetaDharmadhikari:
image
image
image
image

### DIFF
--- a/src/shweta_Dharmadhikari/string_count/FindMinimumWordLength.java
+++ b/src/shweta_Dharmadhikari/string_count/FindMinimumWordLength.java
@@ -1,0 +1,26 @@
+package shweta_Dharmadhikari.string_count;
+
+import java.util.Scanner;
+
+public class FindMinimumWordLength {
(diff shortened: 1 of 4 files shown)


src/shweta_Dharmadhikari/string_count/FindMinimumWordLength.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class FindMinimumWordLength {
+	 String getMaxWordFromString() {
+		 Scanner scan = new Scanner(System.in);
+		 System.out.println("Enter String : ");
+		 String input = scan.nextLine();
+		 String output[] = input.split(" ");
+		 String minWord = output[0];
+		 
+		for (int index = 0; index < output.length;index++) {
```
harshvegada, line 13: As you take smallestString as splitString[0] then the loop can start from 1

harshvegada (approved):

PR #1193, opened 2021-08-21
TanmoyRoy_Assignment20

tanmoyroy96:
Output Program 1: find frequency of given character from user defined string.
image

Output program 2: Find frequency of each character from user defined string.
image

Output program 3: print unique characters from user defined string.
image

Output program 4 : print duplicate characters with their frequency from user defined string.
image


harshvegada (approved):

Take a look at below code
Amol: https://github.com/KrishnaKTechnocredits/JAVATechnoJuly2021/pull/1184/files

PR #1198, opened 2021-08-21
Shantanu_Assignment_20

ShantanuGTechnocredits:
image
image
image
image

### DIFF
--- a/src/shantanu/assignment20/DuplicateCharactersWithFrequency.java
+++ b/src/shantanu/assignment20/DuplicateCharactersWithFrequency.java
@@ -1,0 +1,32 @@
+package shantanu.assignment20;
+
+import java.util.Scanner;
+
+public class DuplicateCharactersWithFrequency {
+	int getFrequencyOfCharacter(String input, char ch) {
+		int count = 0;
+		for(int index = 0; index < input.length(); index++) {
+			if(input.charAt(index) == ch)
+				count++;
+		}
+		return count;
+	}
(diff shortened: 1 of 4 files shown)


src/shantanu/assignment20/DuplicateCharactersWithFrequency.java
```diff
@@ -0,0 +1,32 @@
+package shantanu.assignment20;
+
```
nehatawari, line 2: Write down the problem statement

src/shantanu/assignment20/UniqueCharactersFromString.java
```diff
@@ -0,0 +1,21 @@
+package shantanu.assignment20;
+
+import java.util.Scanner;
+
+public class UniqueCharactersFromString {
+	void displayUniqueCharacters(String input) {
+		System.out.println("Unique characters:");
+		for(int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
+			if(index == input.indexOf(ch))
```
nehatawari, line 10: Condition is wrong it should be
if(input.indexOf(ch)==input.lastIndexOf(ch))

ShantanuGTechnocredits: Corrected the condition and committed the changes with commit message - "Shantanu_Assignment20_Refactoring"

nehatawari (approved):

PR #1201, opened 2021-08-21
Shreya_Assignment_19

ShreyaGulhane:
image

image

image

image

### DIFF
--- a/src/shreya_assignment_19/MaxWordLength.java
+++ b/src/shreya_assignment_19/MaxWordLength.java
@@ -1,0 +1,19 @@
+package shreya_assignment_19;
+
+public class MaxWordLength {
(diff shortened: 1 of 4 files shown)


src/shreya_assignment_19/MaxWordLength.java
```diff
@@ -0,0 +1,19 @@
+package shreya_assignment_19;
+
+public class MaxWordLength {
+	void getMaxWordlength(String str) {
+		String[] arr=str.split(" ");
+		String maxWordlength=arr[0];
+			for(int index=0;index<arr.length;index++) {
```
harshvegada, line 7: As you take smallestString as splitString[0] then the loop can start from 1

harshvegada (approved):

PR #1216, opened 2021-08-21
Purshottam_Assignment20

purshottam94:
1):
image
2)
image
3)
image
4)
image

### DIFF
--- a/src/purshottamJoshi/Assignment20_3.java
+++ b/src/purshottamJoshi/Assignment20_3.java
@@ -1,0 +1,26 @@
+package purshottamJoshi;
+
+class Assignment20_3{
+	void printFreqOfEachChar(String input) {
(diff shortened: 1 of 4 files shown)


src/purshottamJoshi/Assignment20_3.java
```diff
@@ -0,0 +1,26 @@
+package purshottamJoshi;
+
+class Assignment20_3{
+	void printFreqOfEachChar(String input) {
+		System.out.println("Unique characters in a string : ");
+		for(int index=0; index <input.length();index++) {
+			int count =0;
```
surabhi-9, line 7: function which return count of each frequency can called whenever we needed. here we need to call only that method instead of writing whole logic

surabhi-9 (approved):
looks good

PR #1218, opened 2021-08-22
Husain Assignment 21

husainbolt:
image

### DIFF
--- a/src/husain/Assignment21/AgeDifference.java
+++ b/src/husain/Assignment21/AgeDifference.java
@@ -1,0 +1,34 @@
+package husain.Assignment21;
+
+import java.util.Arrays;
+
+public class AgeDifference {
+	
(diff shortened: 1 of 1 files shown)


src/husain/Assignment21/AgeDifference.java
```diff
@@ -0,0 +1,34 @@
+package husain.Assignment21;
+
+import java.util.Arrays;
+
+public class AgeDifference {
+	
+	int findAgeDifference(int arr[]) {
+		int max = 0;
+		int min = arr[0];
+		for(int index = 0; index<arr.length; index++) {
```
harshvegada, line 10: **`Code logic looks good to me!`**
if u took minNumber as input[0] then u should start with index 1, as when u compare index 0 internally both are same think on this.

harshvegada (approved):

PR #1220, opened 2021-08-22
ParthavD_Assignment_21

parthavd:
image


src/parthav/parthavD_Assignment_21/FindDifferenceBetweenOldestAndYoungest.java
```diff
@@ -0,0 +1,27 @@
+package parthav.parthavD_Assignment_21;
+
+public class FindDifferenceBetweenOldestAndYoungest {
+
+	void printDifferenceBetweenOldestYoungest(int[] inputAge) {
+		int maxNumber = 0;
+		int minNumber = inputAge[0];
```
harshvegada, line 7: if u took minNumber as input[0] then u should start with index 1, as when u compare index 0 internally both are same think on this.

harshvegada (approved):

PR #1223, opened 2021-08-22
Precilla_Assignment21

precillaaranha16:
image

### DIFF
--- a/src/precilla/assignment21/AssignmentProgram.java
+++ b/src/precilla/assignment21/AssignmentProgram.java
@@ -1,0 +1,27 @@
+package precilla.assignment21;
+
+import java.util.Arrays;
+public class AssignmentProgram {
+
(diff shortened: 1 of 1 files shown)


src/precilla/assignment21/AssignmentProgram.java
```diff
@@ -0,0 +1,27 @@
+package precilla.assignment21;
+
+import java.util.Arrays;
+public class AssignmentProgram {
+
+	int differenceBetwenOldestAndYougest(int[] array) {
+		int youngestMember=array[0];
+		int oldestMember=array[0];
+		int difference;
+		for(int index=0;index<array.length;index++) {
```
harshvegada, line 10: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1225, opened 2021-08-22
Margi_Assignment_21

MargiSantoki:
image

### DIFF
--- a/src/margi/Assignment_21/Assignment_21.java
+++ b/src/margi/Assignment_21/Assignment_21.java
@@ -1,0 +1,28 @@
+package margi.Assignment_21;
+
+public class Assignment_21 {
+	
+	//difference between oldest and youngest family member
+	int processData(int[] age) {
+		int min=age[0];
+		int max=age[0];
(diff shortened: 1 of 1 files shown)


src/margi/Assignment_21/Assignment_21.java
```diff
@@ -0,0 +1,28 @@
+package margi.Assignment_21;
+
+public class Assignment_21 {
+	
+	//difference between oldest and youngest family member
+	int processData(int[] age) {
+		int min=age[0];
+		int max=age[0];
```
harshvegada, line 8: this should be like

	int min=age[0];
		int max=age[1];

src/margi/Assignment_21/Assignment_21.java
```diff
@@ -0,0 +1,28 @@
...
+		for(int index = 0; index < age.length; index++) {
```
harshvegada, line 9: as both start from 0 n 1 then loop should starts from 2

harshvegada (approved):

PR #1226, opened 2021-08-22
BhaktiK Assignment 21

bhaktibk:
image

### DIFF
--- a/src/bhakti/Assignment_21.java
+++ b/src/bhakti/Assignment_21.java
@@ -1,0 +1,30 @@
+
+/*Assignment - 21 : 22nd Aug'2021
+
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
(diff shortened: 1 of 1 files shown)


src/bhakti/Assignment_21.java
```diff
@@ -0,0 +1,30 @@
...
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
+
+package bhakti;
+
+public class Assignment_21 {
+	 void differenceOldestYoungest(int[] age) {
+		 int lowerAge=age[0];
```
harshvegada, line 13: if u took min/max as age[0] then u should start with index 1, as when u compare index 0 internally both are the same think on this.

harshvegada (approved):

PR #1227, opened 2021-08-22
Purshottam_Assignment21

purshottam94:
image

### DIFF
--- a/src/purshottamJoshi/Assignment21.java
+++ b/src/purshottamJoshi/Assignment21.java
@@ -1,0 +1,31 @@
+package purshottamJoshi;
+
+class Assignment21{
+
+	int ageDifference() {
+		
+		int maxAge=0;
+		int minAge=0;
+		int[] age = {10,34,38,68,72,95,6};
+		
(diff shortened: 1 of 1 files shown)


src/purshottamJoshi/Assignment21.java
```diff
@@ -0,0 +1,29 @@
+package purshottamJoshi;
+
+class Assignment21{
+
+	int ageDifference() {
+		
+		int maxAge=0;
+		int[] age = {10,34,38,68,72,95,6};
+		int minAge = age[0];
```
harshvegada, line 9: if u took min/max as age[0] then u should start with index 1, as when u compare index 0 internally both are the same think on this.

src/purshottamJoshi/Assignment21.java
```diff
@@ -0,0 +1,29 @@
+package purshottamJoshi;
+
+class Assignment21{
+
+	int ageDifference() {
```
harshvegada, line 5: method should be parametrised

harshvegada (approved):

PR #1231, opened 2021-08-22
Devendra_Assignment_21

devendrachandankhede:
image


src/amol_Pawar/assignment21/Assignment_21.java
```diff
@@ -0,0 +1,30 @@
+package amol_Pawar.assignment21;
+
+public class Assignment_21 {
+
+	void getAgeDifference(int age[]){
+		int young=age[0];
+		int old=age[0];
+		for(int index=0;index<age.length;index++)
```
harshvegada, line 8: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1236, opened 2021-08-22
Madhavi_Assignment_21

Madhavi14C:
image

### DIFF
--- a/src/madhavi/Assignment_21.java
+++ b/src/madhavi/Assignment_21.java
@@ -12,10 +12,10 @@ public class Assignment_21 {
 	}
 
 	int getYoungMember(int[] input) {
-		int minAge = 0;
+		int minAge = input[0];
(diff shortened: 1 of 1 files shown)


src/madhavi/Assignment_21.java
```diff
@@ -0,0 +1,35 @@
...
+			if (input[i] > maxAge) {
+				maxAge = input[i];
+			}
+		}
+		return maxAge;
+	}
+
+	int getYoungMember(int[] input) {
+		int minAge = input[0];
+		for (int i = 0; i < input.length; i++) {
```
harshvegada, line 16: **`Code logic looks good to me!`**
if u took min/max as age[0] then u should start with index 1, as when u compare index 0 internally both are the same think on this.

harshvegada (approved):

PR #1237, opened 2021-08-22
Umakant_Assignment_21

Umakant0402:
image

### DIFF
--- a/src/umakant/Assignment21.java
+++ b/src/umakant/Assignment21.java
@@ -1,0 +1,30 @@
+/*
+ 
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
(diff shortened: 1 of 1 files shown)


src/umakant/Assignment21.java
```diff
@@ -0,0 +1,30 @@
...
+output : 89
+ */
+
+package umakant;
+
+public class Assignment21 {
+
+	int getAgeDifferenceBetweenOldestAndYoungest(int[] input) {
+		int maxAge = input[0], minAge = input[0];
+		for (int index = 1; index < input.length; index++) {
```
harshvegada, line 15: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1238, opened 2021-08-22
Nishi assignment 21

NishiBhartii:
image

### DIFF
--- a/src/nishi/Assignment21.java
+++ b/src/nishi/Assignment21.java
@@ -8,7 +8,7 @@ public class Assignment21 {
 	{
 		int max=input[0];
 		int min=input[0];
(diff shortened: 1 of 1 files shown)


src/nishi/Assignment21.java
```diff
@@ -0,0 +1,47 @@
+package nishi;
+
+import java.util.Scanner;
+
+public class Assignment21 {
+	
+	int getDifferenceOfMaxAndMinAge(int[] input)
+	{
+		int max=input[0];
+		int min=input[0];
```
harshvegada, line 10: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1241, opened 2021-08-22
Ritabrata_Assignment21

Ritabrata1996:
image

### DIFF
--- a/src/ritabrata/Assignment21.java
+++ b/src/ritabrata/Assignment21.java
@@ -1,0 +1,48 @@
+/*Assignment - 21 : 22nd Aug'2021
+
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
+
+package ritabrata;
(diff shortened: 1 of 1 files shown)


src/ritabrata/Assignment21.java
```diff
@@ -0,0 +1,48 @@
...
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
+
+package ritabrata;
+import java.util.Scanner;
+public class Assignment21 {
+	int getAgeDif(int[] age) {
+		int maxAge=age[0];
```
surabhi-9, line 12: ideally we should use two different methods to get maxAge and minAge. so we cant cosider first element is min and max, it will either minimum or maximum, so to get these two values, we should write two methods which will return min age from 1 and max age from 2nd method then use difference logic

surabhi-9 (approved):
please rewrite the change suggested and raise PR again with corrected logic

surabhi-9:
merging for now

PR #1244, opened 2021-08-22
Genius_Assignment_21

genius1312:
image

### DIFF
--- a/src/genius/Assignment_21.java
+++ b/src/genius/Assignment_21.java
@@ -1,0 +1,26 @@
+package genius;
+
+public class Assignment_21 {
+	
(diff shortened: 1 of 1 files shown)


src/genius/Assignment_21.java
```diff
@@ -0,0 +1,26 @@
+package genius;
+
+public class Assignment_21 {
+	
+	int diffOfAgeOfOldestAndYoungerFamilyMember(int[] input) {
+		int maxAge = input[0];
+		int minAge = input[0];
```
harshvegada, line 7: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1245, opened 2021-08-22
Mayur_Assignment_21

MayurSTechnoCredit:
image

### DIFF
--- a/src/mayur/Assignment21.java
+++ b/src/mayur/Assignment21.java
@@ -1,0 +1,34 @@
+package mayur;
+
+import java.util.Arrays;
+
+public class Assignment21 {
+	
+	int findAgeDifference(int arr[]) {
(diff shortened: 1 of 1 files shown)


src/mayur/Assignment21.java
```diff
@@ -0,0 +1,34 @@
+package mayur;
+
+import java.util.Arrays;
+
+public class Assignment21 {
+	
+	int findAgeDifference(int arr[]) {
+		int max = 0;
+		int min = arr[0];
+		for(int index = 0; index<arr.length; index++) {
```
harshvegada, line 10: **`Code logic looks good to me!`**
if u took min/max as age[0] then u should start with index 1, as when u compare index 0 internally both are the same think on this.

harshvegada (approved):

PR #1246, opened 2021-08-22
Monika_Assignment_21

MonikaZankar:
image

### DIFF
--- a/src/monika/Assignment_21.java
+++ b/src/monika/Assignment_21.java
@@ -1,0 +1,28 @@
+/*
+ 
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89
+ */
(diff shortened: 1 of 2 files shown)


src/monika/Assignment_21.java
```diff
@@ -0,0 +1,28 @@
...
+int[] age = {10,34,38,68,72,95,6};
+output : 89
+ */
+package monika;
+
+public class Assignment_21 {
+
+	int getAgeDifferenceBetweenOldestAndYungest(int[] input) {
+		int maxAge=input[0], minAge=input[0];
+		for(int index=1;index<input.length;index++) {
```
harshvegada, line 14: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same think on this.

harshvegada (approved):

PR #1247, opened 2021-08-22
Avinash_Assignment_21

AvinashSPatil:
image

### DIFF
--- a/src/avinash/Assignment21.java
+++ b/src/avinash/Assignment21.java
@@ -1,0 +1,37 @@
+package avinash;
+
+import java.util.Arrays;
+
+/*Assignment - 21 : 22nd Aug'2021
+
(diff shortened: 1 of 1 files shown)


src/avinash/Assignment21.java
```diff
@@ -0,0 +1,37 @@
...
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
+
+public class Assignment21 {
+	int findAgeDifference(int arr[]) {
+		int max = 0;
+		int min = arr[0];
+		for (int index = 0; index < arr.length; index++) {
```
harshvegada, line 16: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1249, opened 2021-08-22
Elizabeth_Assignment_21

Elizmary:
image

### DIFF
--- a/src/elizabeth/Assignment21.java
+++ b/src/elizabeth/Assignment21.java
@@ -1,0 +1,28 @@
+package elizabeth;
+
+class Assignment21{
+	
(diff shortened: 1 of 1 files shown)


src/elizabeth/Assignment21.java
```diff
@@ -0,0 +1,28 @@
+package elizabeth;
+
+class Assignment21{
+	
+	int getAgeDiff(int[] age){
+		int minAge=age[0];
+		int maxAge=age[0];
+		int diff=0;
+		for(int index=0;index<age.length;index++){
```
harshvegada, line 9: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1252, opened 2021-08-22
Shalaka_Assignment19

Shalakaprayag15:
image
image
image
image


src/shalaka/StringArray/SumOfNumbers_Assignment19.java
```diff
@@ -0,0 +1,5 @@
+package shalaka.StringArray;
+
+public class SumOfNumbers_Assignment19 {
+
```
nehatawari, line 4: logic is not there for Sum of number

harshvegada (approved):
pending!
merging with existing

PR #1258, opened 2021-08-23
Ashit_Assignment_21

Ashit46:
image

### DIFF
--- a/src/ashit/Assignment21.java
+++ b/src/ashit/Assignment21.java
@@ -1,0 +1,26 @@
+package ashit;
+public class Assignment21 {
+	int getDiffOfAge(int[] input) {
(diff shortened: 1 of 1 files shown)


src/ashit/Assignment21.java
```diff
@@ -0,0 +1,26 @@
+package ashit;
+public class Assignment21 {
+	int getDiffOfAge(int[] input) {
+		int maxAge=0;
+		int minAge=input[0];
+		for(int index=0;index<input.length;index++) {
+		if(input[index]>maxAge) {
```
harshvegada, line 7: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

PR #1264, opened 2021-08-23
Dattaraj Assignment 21

dattaraj123456:
image

### DIFF
--- a/src/dattaraj/Assignment_21.java
+++ b/src/dattaraj/Assignment_21.java
@@ -1,0 +1,29 @@
+package dattaraj;
+
+public class Assignment_21 {
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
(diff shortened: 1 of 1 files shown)


src/dattaraj/Assignment_21.java
```diff
@@ -0,0 +1,29 @@
...
+		int diff = assign_21.getYoungOldDiff(age);
+		System.out.println("Difference between old & young age is :"+diff);
+	}
+
+	int getYoungOldDiff(int[] age) {
+		// TODO Auto-generated method stub
+		int maxAge =age[0];
+		int minAge=age[0];
+		
+		for(int i=0;i<age.length;i++) {
```
harshvegada, line 18: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1267, opened 2021-08-23
Nasir_Assignment21

nasiruddin16:
image


src/nasir_Assignment_21/DifferenceYoungOld.java
```diff
@@ -0,0 +1,30 @@
...
+int[] age = {10,34,38,68,72,95,6};
+output : 89
+*/
+package nasir_Assignment_21;
+
+public class DifferenceYoungOld {
+	int getDiffrence(int[] arr) {
+		int minAge=arr[0];
+		int maxAge=arr[0];
+		for(int index=0;index<arr.length;index++) {
```
harshvegada, line 12: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1268, opened 2021-08-23
Prerana_Assignment_18

preranabirari30:
image

### DIFF
--- a/src/prerana/Assignment18.java
+++ b/src/prerana/Assignment18.java
@@ -1,0 +1,52 @@
+package prerana;
+
+import java.util.Arrays;
+
+public class Assignment18 {
+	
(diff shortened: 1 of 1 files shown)


src/prerana/Assignment18.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+	
+	String[] reverseAtReversePosition(String[] input)
+	{
+		String[] reverse = new String[input.length];
+		for(int i=0;i<input.length;i++)
+		{
+			input[i]=reverseString(input[i]);			
+		}
+		for(int j=0;j<input.length;j++)
```
harshvegada, line 23: you don't require two loop

you should use counter in reverse array variable.

harshvegada (approved):

PR #1271, opened 2021-08-23
AmolVyas_Assignment_21

amolvyas2311:
image

### DIFF
--- a/src/amol_Vyas/Assignment_21.java
+++ b/src/amol_Vyas/Assignment_21.java
@@ -1,0 +1,31 @@
+/*Assignment - 21 : 22nd Aug'2021
+
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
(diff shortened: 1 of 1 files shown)


src/amol_Vyas/Assignment_21.java
```diff
@@ -0,0 +1,31 @@
...
+output : 89*/
+
+package amol_Vyas;
+
+public class Assignment_21 {
+
+	int getDifferenceBetweenOldYoung(int[] input) {
+		int oldest = input[0];
+		int youngest = input[0];
+		for (int index = 0; index < input.length; index++) {
```
harshvegada, line 15: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1278, opened 2021-08-23
Shreya_Assignment_20

ShreyaGulhane:
image

image

image

image


harshvegada (approved):
creat method parameterized so that can be reuse

PR #1279, opened 2021-08-23
Ramkrishna_Assignment_21

ramunikam:
Screenshot 2021-08-23 at 9 55 46 PM

### DIFF
--- a/src/ramkrishna/FamilyMember.java
+++ b/src/ramkrishna/FamilyMember.java
@@ -1,0 +1,30 @@
+package ramkrishna;
+
+public class FamilyMember {
+
+	
+	int gitDifferenceBetweenOldYoung(int[] num) {
+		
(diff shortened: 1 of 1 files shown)


src/ramkrishna/FamilyMember.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class FamilyMember {
+
+	
+	int gitDifferenceBetweenOldYoung(int[] num) {
+		
+		int oldest= num[0];
+		int youngest = num [0];
+	
+		for(int index=0; index<num.length ; index++ ) {
```
harshvegada, line 11: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1280, opened 2021-08-23
Astha_Assignment_21

AsthaPrakash:
image

### DIFF
--- a/src/astha/Assignment21/AgeDifference.java
+++ b/src/astha/Assignment21/AgeDifference.java
@@ -1,0 +1,34 @@
+package astha.Assignment21;
+
+import java.util.Scanner;
+
+public class AgeDifference {
+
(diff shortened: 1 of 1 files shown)


src/astha/Assignment21/AgeDifference.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Scanner;
+
+public class AgeDifference {
+
+	static int getAgeDifference(int [] inputAge) {
+		int maxAge = inputAge[0];
+		int minAge = inputAge[0];
+		
+		for(int index=0; index< inputAge.length; index++) {
```
harshvegada, line 11: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1284, opened 2021-08-23
Shreya assignment 21

ShreyaGulhane:
image


src/shreya_assignment_21/AgeDifference.java
```diff
@@ -0,0 +1,24 @@
+//Find the difference between oldest and youngest family member.
+package shreya_assignment_21;
+public class AgeDifference {
+	int getAgeDifference(int[] age) {
+		int maxAge = age[0], minAge = age[0];
+		for(int index = 0; index < age.length; index++) {
```
harshvegada, line 6: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1286, opened 2021-08-23
Arti_G_Assignment_21

Arti-Saini:
image

### DIFF
--- a/src/arti_G_Assignment_21/DifferenceBwOlderAndYoungest.java
+++ b/src/arti_G_Assignment_21/DifferenceBwOlderAndYoungest.java
@@ -1,0 +1,30 @@
+package arti_G_Assignment_21;
+
+/*Assignment - 21 : 22nd Aug'2021
+
+Find the difference between oldest and youngest family member.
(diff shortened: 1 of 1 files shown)


src/arti_G_Assignment_21/DifferenceBwOlderAndYoungest.java
```diff
@@ -0,0 +1,30 @@
...
+Find the difference between oldest and youngest family member.
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
+public class DifferenceBwOlderAndYoungest {
+
+	int getDifferenceBetweenOldYoungFamilyMember(int[] input) {
+		int oldest = input[0];
+		int youngest = input[0];
+		for (int index = 0; index < input.length; index++) {
```
harshvegada, line 14: if u took min/max as age[0]/age[1] then u should start with index 2, as when u compare index 0/1 internally both are the same. think on this.

harshvegada (approved):

PR #1289, opened 2021-08-23
ankita_assignment_21

Avkene:
image

### DIFF
--- a/src/ankita_assignment_21/Agedifference.java
+++ b/src/ankita_assignment_21/Agedifference.java
@@ -1,0 +1,23 @@
+package ankita_assignment_21;
+
+public class Agedifference {
+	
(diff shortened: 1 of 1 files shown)


src/ankita_assignment_21/Agedifference.java
```diff
@@ -0,0 +1,23 @@
+package ankita_assignment_21;
+
+public class Agedifference {
+	
+	int diffAge(int [] arr) {
+		int max=arr[0];
+		int min=arr[0];
+		for(int index=0;index<arr.length;index++) {
```
harshvegada, line 8: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1292, opened 2021-08-23
Indrani_Assignment_21

Indrani0201:
image

### DIFF
--- a/src/indrani/Assignment_21.java
+++ b/src/indrani/Assignment_21.java
@@ -1,0 +1,22 @@
+package indrani;
+
+public class Assignment_21{
+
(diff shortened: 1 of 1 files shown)


src/indrani/Assignment_21.java
```diff
@@ -0,0 +1,22 @@
+package indrani;
+
+public class Assignment_21{
+
+	int getAgeDifferenceBetweenOldestAndYoungest(int[] input){
+		int maximumAge = input[0];
+		int minimumAge = input[0];
+		
+		for (int index = 1; index < input.length; index++){
```
harshvegada, line 9: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1295, opened 2021-08-23
Deepak_Assignment_21

dkankhar:
Program 1:
image

### DIFF
--- a/src/deepak/Assignment_21/FamilyMemAge.java
+++ b/src/deepak/Assignment_21/FamilyMemAge.java
@@ -1,0 +1,31 @@
+/*Assignment 21 [Find the difference between oldest and youngest family member.] DATE 08/22
+Age is in the form of array.
+int[] age = {10,34,38,68,72,95,6};
+output : 89*/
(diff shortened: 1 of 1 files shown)


src/deepak/Assignment_21/FamilyMemAge.java
```diff
@@ -0,0 +1,31 @@
...
+output : 89*/
+
+package deepak.Assignment_21;
+
+public class FamilyMemAge {
+
+	int getAgeDifference(int[] arr) {
+		int maxAge = arr[0];
+		int minAge = arr[0];
+		for (int i = 0; i < arr.length; i++) {
```
harshvegada, line 13: **`Code logic looks good to me!`**

harshvegada (approved):

src/deepak/Assignment_21/FamilyMemAge.java
dkankhar: Okay, Thanks Harsh.

PR #1299, opened 2021-08-23
Indu_Assignment_21

InduBhartiya:
image

### DIFF
--- a/src/indu/Assignment_21.java
+++ b/src/indu/Assignment_21.java
@@ -1,0 +1,34 @@
+package indu;
+
+import java.util.Arrays;
+
+public class Assignment_21 {
+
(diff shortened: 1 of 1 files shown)


src/indu/Assignment_21.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Arrays;
+
+public class Assignment_21 {
+
+	int  findDiffrenceBtnOldestandYoungest(int[] age)
+	{
+		int max =0;
+		int min = age[0];
+		for(int index=0 ; index  < age.length ;index++)
```
harshvegada, line 11: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1302, opened 2021-08-24
Monali_Assignment_21

Monaliborole:
image


src/monali/Assignment_21/OldestAndYoungestDiff.java
```diff
@@ -0,0 +1,24 @@
+package monali.Assignment_21;
+
+public class OldestAndYoungestDiff {
+
+	int findDiffOldandYoungMember(int[] age) {
+		int oldest = age[0];
+		int youngest = age[0];
+		for (int index = 0; index < age.length; index++) {
```
harshvegada, line 8: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1307, opened 2021-08-24
Kalpesh_Assignment_21

get-kalpesh:
image

### DIFF
--- a/src/kalpesh/Assignment_21.java
+++ b/src/kalpesh/Assignment_21.java
@@ -1,0 +1,29 @@
+package kalpesh;
+
+public class Assignment_21 {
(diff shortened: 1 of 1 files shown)


src/kalpesh/Assignment_21.java
```diff
@@ -0,0 +1,29 @@
+package kalpesh;
+
+public class Assignment_21 {
+	int getDifferenceBetweenOldYoung(int[] input) {
+		int oldest = input[0];
+		int youngest = input[0];
+		for (int index = 0; index < input.length; index++) {
```
harshvegada, line 7: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1312, opened 2021-08-24
Vivek_Assignment_21

vivekvitthalpatil:
image

### DIFF
--- a/src/vivek/Assignment21.java
+++ b/src/vivek/Assignment21.java
@@ -1,0 +1,22 @@
+package vivek;
+
+public class Assignment21 {
+
+	int diffOldestYoungest(int[] age) {
+		int maxAge = age[0];
+		int minAge = age[0];
+		for (int index = 0; index < age.length; index++) {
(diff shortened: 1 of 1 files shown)


src/vivek/Assignment21.java
```diff
@@ -0,0 +1,22 @@
+package vivek;
+
+public class Assignment21 {
+
+	int diffOldestYoungest(int[] age) {
+		int maxAge = age[0];
+		int minAge = age[0];
+		for (int index = 0; index < age.length; index++) {
```
harshvegada, line 8: **`Code logic looks good to me!`**

harshvegada (approved):

src/vivek/Assignment21.java
vivekvitthalpatil: Thank you for your feedback Harsh.
Agreed and noted the point. At index 0 we are comparing the no with itself so starting from the next index makes more sense.

PR #1314, opened 2021-08-24
Santosh_Assignment_20

Santoshghige:
image


nehatawari:
Use meaningful variable name. don't use i ,c etc

PR #1318, opened 2021-08-24
Sourabh_Assignment_21

sourabhskulkarni:
image

### DIFF
--- a/src/sourabh/Assignment_21.java
+++ b/src/sourabh/Assignment_21.java
@@ -1,0 +1,24 @@
+package sourabh;
+
+public class Assignment_21 {     //Difference Between Older And Youngest
+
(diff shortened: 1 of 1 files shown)


src/sourabh/Assignment_21.java
```diff
@@ -0,0 +1,24 @@
+package sourabh;
+
+public class Assignment_21 {     //Difference Between Older And Youngest
+
+	int getDifferenceBetweenOldYoungFamilyMember(int[] input) {
+		int oldest = input[0];
+		int youngest = input[0];
+		for (int index = 0; index < input.length; index++) {
```
harshvegada, line 8: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1324, opened 2021-08-24
Himanshu_Assignment_21

himanshugpt56:
image

### DIFF
--- a/src/himanshu_array_assignment/Assignment_21.java
+++ b/src/himanshu_array_assignment/Assignment_21.java
@@ -1,0 +1,38 @@
+package himanshu_array_assignment;
+
+public class Assignment_21 {
+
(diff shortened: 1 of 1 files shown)


src/himanshu_array_assignment/Assignment_21.java
```diff
@@ -0,0 +1,38 @@
...
+	int oldestMember;
+	int youngestMember;
+	int diffOfBoth;
+
+	void differnceBtwOldestYoungest(int[] age) {
+
+		int older = age[0];
+		int younger = age[0];
+
+		for (int index = 0; index < age.length; index++) {
```
harshvegada, line 14: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1359, opened 2021-08-25
Chirag_Assignment_21

ChiragPrajapati77:
image

### DIFF
--- a/src/chirag/assignment21/AgeDiffrence.java
+++ b/src/chirag/assignment21/AgeDiffrence.java
@@ -1,0 +1,25 @@
+package chirag.assignment21;
+
+public class AgeDiffrence {
+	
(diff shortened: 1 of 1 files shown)


src/chirag/assignment21/AgeDiffrence.java
```diff
@@ -0,0 +1,25 @@
+package chirag.assignment21;
+
+public class AgeDiffrence {
+	
+	int ageDiffrenceBetweenYoungAndOld(int [] input) {
+		
+		int youngestMember = input[0];
+		int oldestMember = input[0];
+		int diffBetweenYoungAndOld;
+		for(int index=0;index < input.length;index++) {
```
harshvegada, line 10: **`Code logic looks good to me!`**

harshvegada (approved):

PR #1364, opened 2021-08-25
Assignment_21

devendrachandankhede:
image


src/devendra_Assignment_21/Assignment21_DiffBetweenOld_Young.java
```diff
@@ -0,0 +1,27 @@
...
+
+//Find Age Difference Between Oldest And Youngest Family Members
+
+public class Assignment21_DiffBetweenOld_Young {
+
+	void printDiffBetweenOld_Young(int[] inputAge) {
+		int max = 0;
+		int min = inputAge[0];
+
+		for (int index = 0; index < inputAge.length; index++) {
```
harshvegada, line 11: **`Code logic looks good to me!`**
if u took min/max as age[0] then u should start with index 1, as when u compare index 0 internally both are the same. think on this.

PR #1477, opened 2021-08-26
Paresh stretegy

paresh129:
image


KrishnaKTechnocredits:
@paresh129 
you not following the branch creation as per steps, Harsh already putted message for the same twice on group, still
Merging directly 

**Watch git videos again**

PR #1507, opened 2021-08-27
Suchita_Assignment_23

shuchkrish:
image

### DIFF
--- a/src/suchita/Assignment23.java
+++ b/src/suchita/Assignment23.java
@@ -1,0 +1,200 @@
+//Assignment 23
+
+import java.util.Scanner;
+class Assignment23 {
+	
+	//int bankIndex=0;
+	int creditAll=0;
+	int debitAll=0;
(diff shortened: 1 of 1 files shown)


src/suchita/Assignment23.java
```diff
@@ -0,0 +1,200 @@
...
+	int debitAll=0;
+	int creditBob=0; int creditHdfc=0; int creditIcici=0; int creditSbi=0;
+	int debitBob=0; int debitHdfc=0; int debitIcici=0; int debitSbi=0; 
+	double balBob=1000; double balHdfc=2000; double balIcici=3000; double balSbi=4000;
+	double bal1=0; double bal2=0; double bal3=0; double bal4=4;
+	
+	void processBank(int operation) {
+		
+		Scanner scanner = new Scanner(System.in);
+		switch(operation) {
```
KrishnaKTechnocredits, line 17: you should never use Scanner class in service method, take all the required input from main method only, n call this method those params.

KrishnaKTechnocredits (approved):
Rest code looks good to me

PR #1509, opened 2021-08-27
Precilla_Assignment23

precillaaranha16:
image

### DIFF
--- a/src/precilla/assignment23/BankAccounts.java
+++ b/src/precilla/assignment23/BankAccounts.java
@@ -1,0 +1,131 @@
+package precilla.assignment23;
+
+public class BankAccounts {
+	String bankName="";
+	static int debitOperation;
+	static int creditOperation;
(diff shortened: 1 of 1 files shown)


src/precilla/assignment23/BankAccounts.java
```diff
@@ -0,0 +1,131 @@
...
+	int iciciDebitCount,sbiDebitCount,bobDebitCount,hdfcDebitcount;
+	int iciciCreditCount,sbiCreditCount,bobCreditCount,hdfcCreditCount;
+	private float iciciBankBalance=0,hdfcBankBalance=0,bobBankBalance=0,sbiBankbalance=0;
+	float total;
+	
+	void performData(String name,float amount,String operation) {
+		bankName=name;
+		switch(bankName){
+			case "ICICI":
+				if(operation=="Debit")
```
KrishnaKTechnocredits, line 16: You should use .equals not ==

KrishnaKTechnocredits (approved):
You should use .equals not ==
Rest code looks good to me

PR #1514, opened 2021-08-27
AkanshaJain_Assignment_23

akansha-jain6:
Assignment_23
Screenshot 2021-08-27 at 8 09 01 PM

### DIFF
--- a/src/akanksha_Jain/Assignment_23.java
+++ b/src/akanksha_Jain/Assignment_23.java
@@ -1,0 +1,125 @@
+package akanksha_Jain;
+
+public class Assignment_23 {
+	static int totalCreditCount, totalDebitCount;
+	static int totalBalance;
+	static int iciciCreditCount, hdfcCreditCount, bobCreditCount, sbiCreditCount;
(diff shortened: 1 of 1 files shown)


src/akanksha_Jain/Assignment_23.java
```diff
@@ -0,0 +1,125 @@
...
+				totalCreditCount ++;
+				break;
+			default:
+				System.out.println("Invalid");
+			}
+			break;
+		case "Debit":
+			switch(bAccountName) {
+			case "ICICI":
+				totalICICIBal -= amt;
```
KrishnaKTechnocredits, line 69: Before debit you should check the main account balance.

KrishnaKTechnocredits (approved):
code looks good to me

src/akanksha_Jain/Assignment_23.java
akansha-jain6: > Before debit you should check the main account balance.

I think I am updating the balance at the same time when I am doing credit/debit.
I didn't understand where it is wrong. I need guidance

PR #1517, opened 2021-08-27
Elizabeth_Assignment_23

Elizmary:
image

### DIFF
--- a/src/elizabeth/Assignment23.java
+++ b/src/elizabeth/Assignment23.java
@@ -1,0 +1,117 @@
+package elizabeth;
+
+class Assignment23{
+	
+	String bankName;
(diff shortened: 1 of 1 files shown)


src/elizabeth/Assignment23.java
```diff
@@ -0,0 +1,117 @@
...
+	String bankName;
+	int iciciBalance,hdfcBalance,bobBalance,sbiBalance,iciciDebitCount,iciciCreditCount,hdfcDebitCount,hdfcCreditCount,bobDebitCount,bobCreditCount,sbiDebitCount,sbiCreditCount;
+	static int totalDebitCount;
+	static int totalCreditCount;
+	
+	void getBankName(int choice, String operation,int amount){
+		switch (choice){
+			case 1: 
+			bankName="ICICI";
+				if(operation=="credit"){
```
KrishnaKTechnocredits, line 14: you should use .equals method not ==

KrishnaKTechnocredits (approved):
You should use .equals not ==
Rest code looks good to me

src/elizabeth/Assignment23.java
Elizmary: oh yeah totally missed this. Will surely implement this next time.

PR #1520, opened 2021-08-27
Vivek_Assignment_23

vivekvitthalpatil:
image

### DIFF
--- a/src/vivek/Assignment23.java
+++ b/src/vivek/Assignment23.java
@@ -1,0 +1,115 @@
+package vivek;
+
+public class Assignment23 {
+
+	int balanceICICI = 10000, balanceHDFC = 5000, balanceBOB = 2000, balanceSBI = 1000;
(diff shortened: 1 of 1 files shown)


src/vivek/Assignment23.java
```diff
@@ -0,0 +1,115 @@
...
+			}
+			break;
+		case "SBI":
+			switch (transaction) {
+			case "Credit":
+				balanceSBI = balanceSBI + amount;
+				creditSBI++;
+				break;
+			case "Debit":
+				balanceSBI = balanceSBI - amount;
```
KrishnaKTechnocredits, line 54: Before debit you should check the main account balance.

KrishnaKTechnocredits (approved):
Rest code looks good to me

src/vivek/Assignment23.java
vivekvitthalpatil: Noted krishna.

PR #1523, opened 2021-08-27
bank program using switch case

gaurav-4uk:


src/gauravk/Assignment_23/BanksSummaryOperationsAndStatus.java
```diff
@@ -0,0 +1,218 @@
...
+	
+	void operationMenu(String name) {
+		System.out.println("Please select the bank by typing a letter against banks :");
+		int i=1;
+		do {
+			System.out.println("I --> ICIC bank\nH --> HDFC bank\nB --> BOB bank\nS --> SBI bank\nX --> Exit.");
+			char bankSelection = sc.next().charAt(0);
+			switch(bankSelection) {
+				case 'i':
+				case 'I':
```
KrishnaKTechnocredits, line 54: instead of one letter you should give full name so that it will be easy to understand.

KrishnaKTechnocredits (approved):
Rest code looks good to me

PR #1530, opened 2021-08-27
Ritabrata_Assignment23

Ritabrata1996:
image

### DIFF
--- a/src/ritabrata/Assignment23.java
+++ b/src/ritabrata/Assignment23.java
@@ -1,0 +1,170 @@
+/*Assignment - 23 : 27th Aug'2021
+
+A person has total 4 accounts in ICICI, HDFC, BOB, SBI one in each.
+          a) Person can debit or credit amount from respective bank.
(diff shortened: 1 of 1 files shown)


src/ritabrata/Assignment23.java
```diff
@@ -0,0 +1,170 @@
...
+  
+public class Assignment23 {
+	String bankName;
+	int balICICI=1000, balHDFC=1000, balBOB=1000, balSBI=1000;
+	int debitCountICICI=0, debitCountHDFC=0, debitCountBOB=0, debitCountSBI=0;
+	int creditCountICICI=0, creditCountHDFC=0, creditCountBOB=0, creditCountSBI=0;
+	static int totalCrCnt, totalDrCnt;
+	void bankTxn(int choice, String optrn, int amt) {
+		switch(choice) {
+		case 1:{
```
KrishnaKTechnocredits, line 37: case not require {}

KrishnaKTechnocredits (approved):
You should use .equals not ==
Rest code looks good to me

PR #1549, opened 2021-08-28
Sayaliv_Assignment_23

sayalivt:
image

### DIFF
--- a/src/sayali_V/Assignment_23.java
+++ b/src/sayali_V/Assignment_23.java
@@ -1,0 +1,140 @@
+package sayali_V;
+
+public class Assignment_23 {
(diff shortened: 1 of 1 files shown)


src/sayali_V/Assignment_23.java
```diff
@@ -0,0 +1,140 @@
...
+
+		case "ICICI":
+			switch (opt) {
+			case "credit":
+				currantBalanceICICI = currantBalanceICICI + amount;
+				iciciCreditcount++;
+				allCreditCount++;
+				break;
+			case "debit":
+				currantBalanceICICI = currantBalanceICICI - amount;
```
nehatawari, line 37: try to handle negative scenarios like if balance is less than amout.

PR #1583, opened 2021-08-29
Manaswi_Assignment_23

ManaswiHumbe:
image

### DIFF
--- a/src/manaswi/Assignment_23.java
+++ b/src/manaswi/Assignment_23.java
@@ -1,0 +1,115 @@
+package manaswi;
+
+public class Assignment_23 
+{
+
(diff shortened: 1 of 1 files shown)


src/manaswi/Assignment_23.java
```diff
@@ -0,0 +1,115 @@
...
+	{
+
+		switch(bankName)
+		{
+
+		case "ICICI":
+			switch(operation) 
+			{
+			case "Debit":
+				iciciBalance=iciciBalance-ammount;
```
harshvegada, line 21: before debit you should check main balance

harshvegada (approved):

PR #1597, opened 2021-08-29
Shalaka_Assignment25

Shalakaprayag15:
image
image

### DIFF
--- a/src/shalaka/Assignment25/SumOFNumbers.java
+++ b/src/shalaka/Assignment25/SumOFNumbers.java
@@ -1,0 +1,34 @@
+/*Program 1:
+String str = "te12ch33no3credit4s"; 
+output = 12+33+3+4
+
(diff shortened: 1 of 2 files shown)


src/shalaka/Assignment25/SumOFNumbers.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class SumOFNumbers {
+	int getSumOfNumbers(String input) {
+		int sum = 0;
+		String sumDigit = "0";
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
+			if (Character.isDigit(ch)) {
+				sumDigit += ch;
+			} else {
```
nehatawari, line 18: instead of else condition it should be
else if(!sumDigit.equals(""))

PR #1627, opened 2021-08-29
Chandani_Assignment_25

chandni0604:
Program 1 : 
image

Program 2 : 
image

### DIFF
--- a/src/chandani_Assignment_25/Assignment25a.java
+++ b/src/chandani_Assignment_25/Assignment25a.java
@@ -1,0 +1,51 @@
+/* Assignment 25 : 29th Aug'2021
+
+Program 2:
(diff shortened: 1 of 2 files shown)


src/chandani_Assignment_25/Assignment25a.java
```diff
@@ -0,0 +1,51 @@
...
+output = 12+33+3+4
+
+Program 1:
+String str = "Ma1U2liK"
+output : MUK12ali */
+package chandani_Assignment_25;
+
+public class Assignment25a{
+
+	void printUpperCase(String str) {
```
harshvegada, line 14: just FYI, all these methods can be merge in one n you can handle it with if else condition

harshvegada (approved):

PR #1635, opened 2021-08-29
Nidhi_Assignment_24

Nidy-C:
image

### DIFF
--- a/src/nidhi/Assignment_24_Program1.java
+++ b/src/nidhi/Assignment_24_Program1.java
@@ -1,0 +1,55 @@
+package nidhi;
+
+
+class Assignment_24_Program1{
+	
+	String findVowels(String str){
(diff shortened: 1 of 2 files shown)


src/nidhi/Assignment_24_Program1.java
```diff
@@ -0,0 +1,55 @@
...
+
+class Assignment_24_Program1{
+	
+	String findVowels(String str){
+		String vowels = "";
+		
+		for(int i=0;i <str.length();i++){
+			switch (str.charAt(i)){
+				case 'a':
+					if (vowels.indexOf(str.charAt(i)) >=0){
```
harshvegada, line 12: here you don't require if-else condition as you are storing in the same variable so no impact.

harshvegada (approved):

PR #1642, opened 2021-08-30
Nishi_Scoding_Test2

NishiBhartii:
image

### DIFF
--- a/src/nishi/ScodingTest2.java
+++ b/src/nishi/ScodingTest2.java
@@ -3,9 +3,7 @@ package nishi;
 public class ScodingTest2 {
 	String getOutput(String input)
 	{		
-		String str1="";
-		int num=0;
-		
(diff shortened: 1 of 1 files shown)


src/nishi/ScodingTest2.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class ScodingTest2 {
+	String getOutput(String input)
+	{		
+		String str1="";		
+		for(int i=0;i<input.length();i++)
+		{
+			if(Character.isUpperCase(input.charAt(i)))
+			{
+				str1+=(String.valueOf(input.charAt(i))).toLowerCase();
```
harshvegada, line 11: you should apply toLowerCase() on input String so that all characters converted to lower n you don't require put it on this line

harshvegada (approved):

PR #1649, opened 2021-08-30
Shraddha_CodingTest2

shraddhaatal:
image

### DIFF
--- a/src/shraddha/CodingTest/StringSmallToCapitalViceVersa.java
+++ b/src/shraddha/CodingTest/StringSmallToCapitalViceVersa.java
@@ -1,0 +1,33 @@
+/*Programming Test - 2 : 30th Aug 2021
+
+Convert all lowercase to uppercase and uppercase to lowercase in a given string.
+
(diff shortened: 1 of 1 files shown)


src/shraddha/CodingTest/StringSmallToCapitalViceVersa.java
```diff
@@ -0,0 +1,33 @@
...
+		String inputString = "A1ashVi6";
+		System.out.println("Input String is :- " + inputString);
+		System.out.println("After changing case of each character in the given string.. Output string will be like :-"
+				+ newObject.changeCaseOfString(inputString));
+	}
+
+	// method to change case of each character in the given String
+	String changeCaseOfString(String inputString) {
+		String outputstring = "";
+		for (int index = 0; index < inputString.length(); index++) {
```
harshvegada, line 21: you can apply toLowerCase on input String so that u don't require to call it in every condition

harshvegada (approved):

PR #1663, opened 2021-08-30
Dattaraj Assignment 22

dattaraj123456:
image

### DIFF
--- a/src/dattaraj/Coding_TEST2.java
+++ b/src/dattaraj/Coding_TEST2.java
@@ -1,0 +1,31 @@
+package dattaraj;
+
+public class Coding_TEST2 {
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+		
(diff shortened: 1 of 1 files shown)


src/dattaraj/Coding_TEST2.java
```diff
@@ -0,0 +1,31 @@
...
+		System.out.println("New String is as :" +test2.convert_String(str));
+	}
+
+	String convert_String(String str) {
+		String str_new="";
+		
+		// TODO Auto-generated method stub
+		for (int i = 0; i < str.length(); i++) {
+			if(Character.isUpperCase(str.charAt(i))) {
+				str_new=str_new+Character.toLowerCase(str.charAt(i));
```
harshvegada, line 21: Direct you can apply toLowerCase() on input STR so that u don’t require to write here

harshvegada (approved):

PR #1664, opened 2021-08-30
PrecillaAranha_CodingTest_2

precillaaranha16:
image

### DIFF
--- a/src/precilla/codingtest2/StringProgram.java
+++ b/src/precilla/codingtest2/StringProgram.java
@@ -1,0 +1,26 @@
+package precilla.codingtest2;
+
+public class StringProgram {
+
+	void convertValue(String string) {
+		String output="";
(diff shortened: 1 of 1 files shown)


src/precilla/codingtest2/StringProgram.java
```diff
@@ -0,0 +1,26 @@
+package precilla.codingtest2;
+
+public class StringProgram {
+
+	void convertValue(String string) {
+		String output="";
+		for(int index=0;index<string.length();index++) {
+			char ch=string.charAt(index);
+			if(Character.isLetter(ch)&&Character.isUpperCase(ch))
+				output+=Character.toLowerCase(ch);
```
harshvegada, line 10: Direct you can apply toLowerCase() on input STR so that u don’t require to write here

harshvegada (approved):

PR #1673, opened 2021-08-30
Dattaraj Assignment 23

dattaraj123456:
image

### DIFF
--- a/src/dattaraj/Assignment_23.java
+++ b/src/dattaraj/Assignment_23.java
@@ -1,0 +1,122 @@
+package dattaraj;
+
+public class Assignment_23 {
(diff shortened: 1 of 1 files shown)


src/dattaraj/Assignment_23.java
```diff
@@ -0,0 +1,122 @@
...
+				sbiCreditCount ++;
+				totalCreditCount ++;
+				break;
+			default:
+				System.out.println("Invalid");
+			}
+			break;
+		case "Debit":
+			switch(bAccountName) {
+			case "ICICI":
```
harshvegada, line 68: Before debit you should check the main account balance.

harshvegada (approved):

PR #1679, opened 2021-08-30
Arti g assignment 24

Arti-Saini:
image

### DIFF
--- a/src/arti_G_Assignment24/FrequencyCount.java
+++ b/src/arti_G_Assignment24/FrequencyCount.java
@@ -1,0 +1,93 @@
+package arti_G_Assignment24;
+
+import java.util.Scanner;
+
+/*Assignment - 24 : 28th Aug'2021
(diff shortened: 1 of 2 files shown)


src/arti_G_Assignment24/FrequencyCount.java
```diff
@@ -0,0 +1,93 @@
...
+output : Hi -> 3*/
+public class FrequencyCount {
+
+	String findVowels(String str){
+		String vowels = "";
+
+		for(int i=0;i <str.length();i++){
+			switch (str.charAt(i)){
+				case 'a':
+					if (vowels.indexOf(str.charAt(i)) >=0){
```
harshvegada, line 21: this if-else not require reason if you already have a matching case and if the condition check something like this
`t > 0`
`char > number`

harshvegada (approved):
**`Code logic works fine but you should write direct char`**

PR #1699, opened 2021-08-30
Rushikesh_Assignment_25

rushikeshjunewal:
image
image

### DIFF
--- a/src/rushikesh/stringmethods/SumAsString.java
+++ b/src/rushikesh/stringmethods/SumAsString.java
@@ -1,0 +1,29 @@
+package rushikesh.stringmethods;
+
+public class SumAsString {
+	void getStringAsNum(String str) {
+		
+		int twoDigit = 0;
+		int singleDigit = 0;
+		for(int index=0;index<str.length();index++) {
+			char ch=str.charAt(index);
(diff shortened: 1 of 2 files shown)


src/rushikesh/stringmethods/SumAsString.java
```diff
@@ -0,0 +1,29 @@
+package rushikesh.stringmethods;
+
+public class SumAsString {
+	void getStringAsNum(String str) {
+		
+		int twoDigit = 0;
+		int singleDigit = 0;
+		for(int index=0;index<str.length();index++) {
+			char ch=str.charAt(index);
+			if(Character.isDigit(ch) && Character.isDigit(str.charAt(index+1))) {
```
surabhi-9, line 10: will throw runtime error if index will go out of range

surabhi-9 (approved):
ArrangeOrder code looks good to me, another code needs small correction in logic

src/rushikesh/stringmethods/SumAsString.java
rushikeshjunewal: but we've taken the length, so why it'll go out of range

PR #1704, opened 2021-08-30
Purshottam_Programming_Test2

purshottam94:
image

### DIFF
--- a/src/purshottamJoshi/Programming_Test2.java
+++ b/src/purshottamJoshi/Programming_Test2.java
@@ -1,0 +1,28 @@
+package purshottamJoshi;
+
+class Programming_Test2{
+	
+	void upperLower(String input){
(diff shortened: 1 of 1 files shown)


src/purshottamJoshi/Programming_Test2.java
```diff
@@ -0,0 +1,28 @@
+package purshottamJoshi;
+
+class Programming_Test2{
+	
+	void upperLower(String input){
+		String output="";
+		for(int index=0; index < input.length(); index++){
+			char ch = input.charAt(index);
+			if(Character.isLowerCase(ch)){
```
surabhi-9, line 9: nested if is used instead of if else if, advantage to use if else if ladder is condition will be checked only once

surabhi-9 (approved):

PR #1705, opened 2021-08-30
Riten_CodingTest_2

ritenthakkar:
image

### DIFF
--- a/src/riten/CodingTest2.java
+++ b/src/riten/CodingTest2.java
@@ -1,0 +1,28 @@
+package riten;
+
+class CodingTest2 {
+
+	void upperLower(String input) {
+		String result = "";
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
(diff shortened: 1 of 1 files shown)


src/riten/CodingTest2.java
```diff
@@ -0,0 +1,28 @@
...
+
+	void upperLower(String input) {
+		String result = "";
+		for (int index = 0; index < input.length(); index++) {
+			char ch = input.charAt(index);
+			if (Character.isUpperCase(ch)) {
+				ch = Character.toLowerCase(ch);
+			} else if (Character.isLowerCase(ch)) {
+				ch = Character.toUpperCase(ch);
+			} else if (Character.isDigit(ch)) {
```
surabhi-9, line 13: if nothing needs to do in else if block then we dont need to write blank elseif as well

surabhi-9 (approved):
logic looks good

PR #1708, opened 2021-08-30
Ritabrata_Assignment25

Ritabrata1996:
image

### DIFF
--- a/src/ritabrata/Assignment25.java
+++ b/src/ritabrata/Assignment25.java
@@ -1,0 +1,57 @@
+/*Assignment 25 : 29th Aug'2021
+
+Program 1:
+String str = "te12ch33no3credit4s"; 
+output = 12+33+3+4
+
+Program2:
+String str = "Ma1U2liK"
+output : MUK12ali*/
+
+package ritabrata;
+
+public class Assignment25 {
+	void getIntValues(String input) {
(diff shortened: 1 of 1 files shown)


src/ritabrata/Assignment25.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class Assignment25 {
+	void getIntValues(String input) {
+		String output="",digit="";
+		System.out.println("The input string is "+input);
+		for(int index=0; index<input.length();index++) {
+			char ch= input.charAt(index);
+			
+			if(Character.isDigit(ch)) {
+				digit+= String.valueOf(ch);
```
surabhi-9, line 21: String.valueOf method is not mandatory, you can directly convert character if first variable is string then if you add anything , it will become string like for ex.:
String digit = digit +ch

src/ritabrata/Assignment25.java
```diff
@@ -0,0 +1,57 @@
...
+				continue;
+			}
+			if (digit!="") {
```
surabhi-9, line 24: for equality/non equality, you should use equals method of string because == or != compares with memory, 
can use as below:
if(!digit.equals(""))

surabhi-9 (approved):
logic looks good to me

PR #1725, opened 2021-08-30
Gaurav_Assignment_25

gaurav-4uk:


gaurav-4uk:
Program 1 is having output 56 instead of 52.
Please look into the code and advice how to stop first instance of number from totalling.

harshvegada (approved):

PR #1728, opened 2021-08-30
Sayliv assignment 23

Saylivyavhare:
image

### DIFF
--- a/src/sayli_Vyavhare/Bank_SwitchCase.java
+++ b/src/sayli_Vyavhare/Bank_SwitchCase.java
@@ -1,0 +1,152 @@
+/*Assignment - 23 : 
+
+A person has total 4 accounts in ICICI, HDFC, BOB, SBI one in each.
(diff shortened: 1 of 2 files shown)


src/sayli_Vyavhare/Bank_SwitchCase.java
```diff
@@ -0,0 +1,152 @@
...
+			creditCountICICI, debitCountBOB, creditCountBOB, debitCountHDFC, creditCountHDFC, debitCountSBI,
+			creditCountSBI;
+
+	void bankAndTransactions(String bank, String transaction, int amount) {
+		switch (bank) {
+		case "HDFC":
+			switch (transaction) {
+
+			case "Debit":
+				balanceHDFC = balanceHDFC - amount;
```
nehatawari, line 39: try to cover negative scenarios like if Amount is greater than Balance

nehatawari:
looks ok

PR #1751, opened 2021-08-31
Neha_Assignment_24

NehaJTechnocredits:
### Output for finding word with max frequency in string:
image

### Output to find vowels and their count is:
image


harshvegada (approved):

PR #1882, opened 2021-09-03
removing specific element from array

gaurav-4uk:
After removing specific element 14 from array [25, 14, 56, 15, 36, 56, 77, 18, 29, 49]
output is: [25, 56, 15, 36, 56, 77, 18, 29, 49]


harshvegada (approved):

PR #1978, opened 2021-09-04
Margi_Assignment_28

MargiSantoki:
image

image

### DIFF
--- a/src/margi/Assignment_28/Assignment_28_2.java
+++ b/src/margi/Assignment_28/Assignment_28_2.java
@@ -1,0 +1,35 @@
+package margi.Assignment_28;
+
+public class Assignment_28_2 {
+	
+	//nth highest number from array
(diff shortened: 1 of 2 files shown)


harshvegada (approved):
**`Code looks good to me, Good to merge`**
**`Approved`**

src/margi/Assignment_28/Assignment_28_2.java
```diff
@@ -0,0 +1,35 @@
...
+				}
+			}
+			input[temp]=0;
+			outerIndex++;
+		}
+		for(int index=1; index<input.length; index++) {
+			if(secondMax<input[index])
+				secondMax=input[index];
+		}
+		return secondMax;
```
harshvegada, line 26: why secondMax u r returning?

harshvegada (commented):
Looks like u need to return nth highest number instead of second max

PR #1989, opened 2021-09-04
Bhagyashree_Assignment27

BhagyashreeJYeole:
image

### DIFF
--- a/src/bhagyashree/Assignment27.java
+++ b/src/bhagyashree/Assignment27.java
@@ -1,0 +1,44 @@
+/*
+ * WAP to find 2 arrays are identical or not. 
+ */
+package bhagyashree;
+
+public class Assignment27 {
+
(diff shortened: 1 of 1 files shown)


src/bhagyashree/Assignment27.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package bhagyashree;
+
+public class Assignment27 {
+
+	void checkArraysIdentical(int[] array1, int[] array2) {
+		boolean flag = false;
+		if (array1.length == array2.length) {
+			for (int index = 0; index < array1.length; index++) {
+				if (array1[index] == array2[index]) {
```
harshvegada, line 12: a simple way you should put it **!=** condition so that u don't require to write else part, just FYI

harshvegada (approved):
**`Code looks good to me`**

PR #1985, opened 2021-09-04
Suruchi_Assignment_28

suruGIT21:
image

PR #1997, opened 2021-09-04
SayliVyavhare_Assignment_28

Saylivyavhare:
image


harshvegada (approved):
**`Code looks good to me, find 2nd highest is expected check for the other approach`**

PR #2004, opened 2021-09-04
ParthavD_Assignment_28

parthavd:
image


harshvegada (approved):
Code is a bit lengthy so understand, this approach has multiple loop iterations.
check below SS, let me know if u below logic or not

Screenshot 2021-09-06 at 8 09 39 PM

PR #2007, opened 2021-09-05
Rushikesh_Assign_28

rushikeshjunewal:
image
image


surabhi-9 (approved):
please try changing input to see if 55 only coming as second highest, interchange place of 55 and 99

PR #2009, opened 2021-09-05
Priti_Assingment_27

psnilange:
image

### DIFF
--- a/src/pritiInterger_Array_Programme/Assingment_27_1.java
+++ b/src/pritiInterger_Array_Programme/Assingment_27_1.java
@@ -1,0 +1,54 @@
+package pritiInterger_Array_Programme;
+
+public class Assingment_27_1 {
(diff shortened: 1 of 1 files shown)


src/pritiInterger_Array_Programme/Assingment_27_1.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class Assingment_27_1 {
+	void descideIdenticalArrayOrNot(int[] arr1, int [] arr2) {
+		int lenarr1= arr1.length;
+		int lenarr2= arr2.length;
+		int cnt=0;
+		int cnt1=0;
+		if (lenarr1 !=lenarr2)
+			System.out.println("Both Arrays are not identical and as their legth is not same.");
+		else if (lenarr1 ==lenarr2){
```
surabhi-9, line 11: if condition is not required. only else also would work

surabhi-9 (approved):
rest code looks good to me

PR #2014, opened 2021-09-05
Akshay_Assignment_24

akshay0795:
image
image

### DIFF
--- a/src/akshay/FindVowels.java
+++ b/src/akshay/FindVowels.java
@@ -1,0 +1,59 @@
+package akshay;
+
+import java.util.Scanner;
+
+public class FindVowels {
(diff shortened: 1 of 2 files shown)


src/akshay/FindVowels.java
```diff
@@ -0,0 +1,59 @@
...
+					if(u) {
+						System.out.print("u ");
+						u = false;
+					}
+					break;
+			}
+		}
+	}
+	
+	public final static void main(String[] args) {
```
harshvegada, line 52: main methods not require to be final

harshvegada (approved):
**`Code looks good to me`**

PR #2028, opened 2021-09-05
Akshay_Assignment_27

akshay0795:
image

### DIFF
--- a/src/akshay/ArrayMatching.java
+++ b/src/akshay/ArrayMatching.java
@@ -1,0 +1,54 @@
+package akshay;
+
+import java.util.Arrays;
+
+public class ArrayMatching {
+	
+	void getResult(int[] arr1, int[] arr2) {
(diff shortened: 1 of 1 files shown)


src/akshay/ArrayMatching.java
```diff
@@ -0,0 +1,54 @@
...
+	
+	void getResult(int[] arr1, int[] arr2) {
+		boolean b = false;
+		if(arr1.length != arr2.length) {
+			System.out.println("Arrays are not identical");
+		}
+		else {
+			for(int index = 0; index <= arr1.length-1; index++) {
+				if(arr1[index] == arr2[index]) {
+					b = true;
```
harshvegada, line 15: you can replace this condition with != just to avoid else part n bit code improved

harshvegada (approved):
**`Code looks good to me`**

PR #2029, opened 2021-09-05
Trupti_Assignment_28_1

truptimkulkarni:


src/trupti/Assignment_28_1.java
```diff
@@ -0,0 +1,25 @@
...
+	
+	void secondMaxNumber(int[] input){
+		int max=23;
+		int secondMax=10;
+		for(int index=2; index<input.length; index++){
+			if(input[index]>max){
+				secondMax=max;
+				max=input[index];
+			}
+			else if(secondMax<input[index]){
```
harshvegada, line 19: you need to handle one more condition which is below

max != arr[index] with existing one

harshvegada (approved):
**`Code looks good to me`**

PR #2036, opened 2021-09-06
Nasir_Assignment28

nasiruddin16:
image

### DIFF
--- a/src/nasir_Assignment_28/Assignment_28.java
+++ b/src/nasir_Assignment_28/Assignment_28.java
@@ -1,0 +1,40 @@
+package nasir_Assignment_28;
+
+public class Assignment_28 {
+	int getSecondHighestElement(int[] input) {
+		int highestElement = input[0];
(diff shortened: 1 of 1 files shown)


src/nasir_Assignment_28/Assignment_28.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class Assignment_28 {
+	int getSecondHighestElement(int[] input) {
+		int highestElement = input[0];
+		int secondHighestElement = input[1];
+		for (int index = 2; index < input.length; index++) {
+			if (input[index] > highestElement) {
+				secondHighestElement = highestElement;
+				highestElement = input[index];
+			} else if (input[index] > secondHighestElement)
```
harshvegada, line 11: you need to handle one more condition here

max != arr[index]

harshvegada (approved):
Code looks good to me,
just FYI, diff solution

PR #2039, opened 2021-09-06
Assignment_28

aniket0811:
program 1  
image
program 2
image

### DIFF
--- a/src/aniket/Assignment_28_1.java
+++ b/src/aniket/Assignment_28_1.java
@@ -1,0 +1,34 @@
+package aniket;
+
+import java.util.Arrays;
+
+public class Assignment_28_1 {
+	void getSecondHighestNumber(int[] arr) {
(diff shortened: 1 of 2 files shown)


src/aniket/Assignment_28_1.java
```diff
@@ -0,0 +1,34 @@
...
+		if (maxnumber < secondHighestNumber) {
+			maxnumber = arr[1];
+			secondHighestNumber = arr[0];
+
+		}
+		for (int index = 2; index < arr.length; index++) {
+			if (maxnumber < arr[index]) {
+				secondHighestNumber = maxnumber;
+				maxnumber = arr[index];
+			} else if (secondHighestNumber < arr[index]) {
```
harshvegada, line 18: **`Code looks good to me, find 2nd highest is expected check for the other approach`**

harshvegada (approved):

PR #2042, opened 2021-09-06
Himanshu_Assignment_27

himanshugpt56:
image

### DIFF
--- a/src/himanshu_array_assignment/ArraysIdenticalOrNot.java
+++ b/src/himanshu_array_assignment/ArraysIdenticalOrNot.java
@@ -1,0 +1,40 @@
+package himanshu_array_assignment;
+
+public class ArraysIdenticalOrNot {
+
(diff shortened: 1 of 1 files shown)


src/himanshu_array_assignment/ArraysIdenticalOrNot.java
```diff
@@ -0,0 +1,40 @@
+package himanshu_array_assignment;
+
+public class ArraysIdenticalOrNot {
+
+	void compareArrays(int[] arr1, int[] arr2) {
+		boolean status = true;
+		if (arr1.length == arr2.length) {
+			for (int index = 0; index < arr1.length; index++) {
+				if (arr1[index] == arr2[index])
```
harshvegada, line 9: you can write != condition as well, just to avoid else part, just code improvements

harshvegada (approved):
**`Code looks good to me`**

PR #2044, opened 2021-09-06
Assignment27And28

RasikaNitin:
image
image
image

### DIFF
--- a/src/rasika/Assignment27.java
+++ b/src/rasika/Assignment27.java
@@ -1,0 +1,52 @@
+package rasika;
+
+/*WAP to find 2 arrays are identical or not. 
+int[] arr1 = {10,12,55,32,17};
+int[] arr2 = {10,12,55,32,17}};
+output : Identical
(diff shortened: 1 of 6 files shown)


src/rasika/Assignment27.java
```diff
@@ -0,0 +1,52 @@
...
+int[] arr1 = {10,12,55,32,17};
+int[] arr2 = {10,12,99,32,17}};
+output : Not Identical*/
+public class Assignment27 {
+
+	void checkIdenticalArray(int[] arr1, int[] arr2) {
+		boolean flag = false;
+		if (arr1.length == arr2.length) {
+			for (int index = 0; index < arr1.length; index++) {
+				if (arr1[index] == arr2[index]) {
```
harshvegada, line 21: you can use != as well to avoid else part, just code improvements

harshvegada (approved):
**`Code looks good to me`**

PR #2050, opened 2021-09-06
Akshay_Assignment_28

akshay0795:
image
image

### DIFF
--- a/src/akshay/SecondHighestNum.java
+++ b/src/akshay/SecondHighestNum.java
@@ -1,0 +1,29 @@
+package akshay;
+
+import java.util.Arrays;
+
+public class SecondHighestNum {
+	
+	int max = 0;
+	int secondMax = 0;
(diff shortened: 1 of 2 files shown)


src/akshay/SecondHighestNum.java
```diff
@@ -0,0 +1,29 @@
...
+	int max = 0;
+	int secondMax = 0;
+	
+	int getSecondHighestNum(int[] input) {
+		for(int index = 0; index < input.length; index++) {	
+			if(input[index]>max) {
+				secondMax = max;
+				max = input[index];
+			}
+			else if(input[index] > secondMax) {
```
harshvegada, line 16: you should add one more condition which is below one

input[index] > secondMax && max != input[index]

harshvegada (approved):
**`Code looks good to me`**

PR #2057, opened 2021-09-07
Harshada_Assignment_28

HarshadaMandhare:
image
image

### DIFF
--- a/src/harshada/Assignment_28/Assignment_28_01.java
+++ b/src/harshada/Assignment_28/Assignment_28_01.java
@@ -1,0 +1,46 @@
+/*
+Program 1: Find second highest number from array.
+int[] arr = {10,23,2,11,55,43,99};
+output : 55
+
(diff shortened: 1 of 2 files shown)


src/harshada/Assignment_28/Assignment_28_01.java
```diff
@@ -0,0 +1,46 @@
...
+		System.out.println("Input Array elements are: "+ Arrays.toString(inputArray));
+		int max=0;
+		int secondMax=0;
+		for(int index=0; index < inputArray.length; index++) {
+			if(inputArray[index] > max) {
+			secondMax=max;
+			max=inputArray[index];
+		}
+		else 
+			if(inputArray[index]> secondMax ) {
```
harshvegada, line 31: Here you need to add one more condition

inputArray[index]> secondMax && max != inputArr[index]

harshvegada (approved):
**`Code looks good to me,`**

PR #2058, opened 2021-09-07
AmolVyas_Assignment_29

amolvyas2311:
image

### DIFF
--- a/src/amol_Vyas/Assignment_29.java
+++ b/src/amol_Vyas/Assignment_29.java
@@ -1,0 +1,69 @@
+package amol_Vyas;
+
+public class Assignment_29 {
+
+	String getReverseString(String input) {
+		String output = "";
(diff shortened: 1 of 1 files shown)


src/amol_Vyas/Assignment_29.java
```diff
@@ -0,0 +1,69 @@
...
+
+	boolean checkPalindromNumber(int number) {
+		boolean flag = false;
+		int reminder = 0, sum = 0, temp = number;
+		while (number > 0) {
+			reminder = number % 10;
+			sum = (sum * 10) + reminder;
+			number = number / 10;
+		}
+		if (sum == temp)
```
harshvegada, line 44: you should use .equals method which compares the content of string n this == compares the reference

harshvegada (approved):
**`Code looks good to me`**

PR #2073, opened 2021-09-07
finding nth highest number from array

gaurav-4uk:
Input:
int[] arr = {10,23,2,11,55,43,99};
input n = 3;

Output:
The nth highest number from the array [10, 23, 2, 11, 55, 43, 99] is 43, where n=3


harshvegada (approved):
**`Code looks good to me`**

PR #2113, opened 2021-09-08
Prerana_Assignment_27

preranabirari30:
image

### DIFF
--- a/src/prerana/SimilarArray.java
+++ b/src/prerana/SimilarArray.java
@@ -1,0 +1,36 @@
+
+package prerana;
+
+import java.util.Arrays;
+
+public class SimilarArray {
+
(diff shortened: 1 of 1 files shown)


src/prerana/SimilarArray.java
```diff
@@ -0,0 +1,36 @@
+
+package prerana;
+
+import java.util.Arrays;
+
+public class SimilarArray {
+
+	void getsimilarArray(int arr1[], int arr2[]) {
+
+		if(Arrays.equals(arr1, arr2))
```
harshvegada, line 10: we should use write logic, not use predefined methods

harshvegada (approved):
**`Code looks good to me`**

PR #2117, opened 2021-09-08
Neha assignment 29

NehaJTechnocredits:
### String pallindrome Program 1 with Reverse:
image

### String pallindrome  Program 2 without Reverse:
image

### Number Pallindrome Program:
image


harshvegada (approved):
**`Code looks good to me`**

PR #2133, opened 2021-09-08
Purshottam_Assignment30

purshottam94:
1)
image
2)
image

### DIFF
--- a/src/purshottamJoshi/Assignment30.java
+++ b/src/purshottamJoshi/Assignment30.java
@@ -1,0 +1,49 @@
+package purshottamJoshi;
+
+class Assignment30{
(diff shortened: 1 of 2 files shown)


src/purshottamJoshi/Assignment30.java
```diff
@@ -0,0 +1,49 @@
...
+		while (num > 0) {
+			int a;
+			a = num % 10;
+			sum = sum * 10 + a;
+			num = num / 10;
+		}
+		if (sum == num1)
+		
+			return true;
+			return false;
```
harshvegada, line 31: what is this?
this is CE

PR #2176, opened 2021-09-09
AkanshaJain Coding Test 5

akansha-jain6:
Coding Test 5
Screenshot 2021-09-09 at 4 04 34 PM

I dont know why 5th number is not printing here in response might be because its a very large number.


harshvegada (approved):
**`Code logic looks good to me!, Good to merge`**

PR #2217, opened 2021-09-09
Elizabeth_CodingTest_5

Elizmary:
image

5th perfect number is 33550336. May be its taking time in executing the loop. So that is not getting printed.


harshvegada (approved):
**`Code logic looks good to me!, Good to merge`**

PR #2294, opened 2021-09-10
Anuja_Codingtest_7

AnujaTechno:
image

### DIFF
--- a/src/anuja_Test/Test7.java
+++ b/src/anuja_Test/Test7.java
@@ -1,0 +1,26 @@
+package anuja_Test;
+
+public class Test7 {
+	
+	void display(int[] input) {
+		int count=0;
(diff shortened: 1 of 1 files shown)


src/anuja_Test/Test7.java
```diff
@@ -0,0 +1,26 @@
+package anuja_Test;
+
+public class Test7 {
+	
+	void display(int[] input) {
+		int count=0;
+		for(int i=-2;i<input.length-3;i++) {
```
harshvegada, line 7: i should start from 0

if(arr[i]+1 == arr[1] && arr[i]+1 == arr[2])

harshvegada: input < input.length-2

PR #2301, opened 2021-09-10
Purshottam_CodingTest_6

purshottam94:
1)
image
2)
image

### DIFF
--- a/src/purshottamJoshi/CodingTest6_1.java
+++ b/src/purshottamJoshi/CodingTest6_1.java
@@ -1,0 +1,26 @@
+package purshottamJoshi;
+
+class CodingTest6_1{
+	
+	void firstNonRepeating(String str){
+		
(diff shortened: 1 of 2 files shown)


src/purshottamJoshi/CodingTest6_1.java
```diff
@@ -0,0 +1,26 @@
+package purshottamJoshi;
+
+class CodingTest6_1{
+	
+	void firstNonRepeating(String str){
+		
+		for(int index=0; index < str.length(); index++){
+			boolean checkNonRepeatchar=true;
+			for(int innerindex=0; innerindex <str.length(); innerindex++){
```
harshvegada, line 9: you can use index == str.lastIndexOf(word) then first can consider this as Non Repeating Word

harshvegada (approved):

PR #2312, opened 2021-09-10
Santosh_CodingTest_5

Santoshghige:
image


src/santosh/CodingTest_5.java
```diff
@@ -0,0 +1,31 @@
+package santosh;
+public class CodingTest_5 {
+	boolean isPerfectNumber(long input) {
+		long output = 0;
+		for (long index = 1; index < input; index++) {
```
nehatawari, line 5: Try Optimization
Index<=input/2

nehatawari:
Rest logic looks good

PR #2334, opened 2021-09-10
Devendra_CodingTest_6

devendrachandankhede:
image


harshvegada (approved):
you can use index == str.lastIndexOf(word) then first can consider this as Non Repeating Word

PR #2538, opened 2021-09-14
Shantanu_Assignment_34

ShantanuGTechnocredits:
image
image
image

### DIFF
--- a/src/shantanu/assignment34/SumAfterTriplet.java
+++ b/src/shantanu/assignment34/SumAfterTriplet.java
@@ -1,0 +1,46 @@
+package shantanu.assignment34;
+
+import java.util.Arrays;
+
+public class SumAfterTriplet {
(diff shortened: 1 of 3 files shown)


src/shantanu/assignment34/SumAfterTriplet.java
```diff
@@ -0,0 +1,46 @@
...
+				index += 2;
+				outputIndex += 4;
+			}
+			else {
+				output[outputIndex] = input[index];
+				outputIndex++;
+			}
+			
+		}
+		if(index == input.length-1)
```
nehatawari, line 32: condition not satisfied for different input like {1,3,4,5,7,3,9,10,11,14,44,67,1,2,3,99,4};
try this condition
while(index<input.length) {
			output[outputIndex] = input[index];
			index++;
			outputIndex++;
		}

nehatawari:
rest code looks good to me

PR #2577, opened 2021-09-15
Sagar_Assignment31

sagaanna:
image

### DIFF
--- a/src/sagar_Anasane/Assignment31.java
+++ b/src/sagar_Anasane/Assignment31.java
@@ -1,0 +1,106 @@
+/*Assignment - 31 : 9th Sep'2021
+Program 1: verify given number is a perfect number or not.
+input : 6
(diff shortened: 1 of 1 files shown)


src/sagar_Anasane/Assignment31.java
```diff
@@ -0,0 +1,106 @@
...
+			System.out.println(num+" is a Perfect Number");
+		}else
+			System.out.println(num+" is not a Perfect Number");
+		return num;
+	}
+	
+	void givenNoIsPerfectSquare(int num) {
+		int var=0;
+		for(int index=1;index<num;index++) {
+			if(num%index==0) {
```
nehatawari, line 48: Condition can be improve..go through others coding or simply take one Boolean flag n codition like if(index*index=num)set flag to true

nehatawari:
Rest code looks good

PR #2588, opened 2021-09-16
Neha_Assignment_33

NehaJTechnocredits:
### Program 1 Output : shift all non zeros at the end
image

### Program 2 Output: place all negative numbers at the beginning and all positive numbers at the end
image

### program 3 Output: place all negative numbers at the beginning, zeros and all positive numbers at the end
image

### program 4 Output : find the triplet which in sequence and having maximum sum
image


harshvegada:
**`Approved`**

PR #2630, opened 2021-09-18
NehaJ_CodingTest_6

NehaJTechnocredits:
### Program 1: Find first non repeating character from a word. [expected : 10 mins, actual time : 5 min]
image

### Program 2: Find first non repeating word from a given String. [expected : 15 mins, actual = 10 min]
image


harshvegada:
**`Approved`**

PR #2998, opened 2021-10-01
Finding length of string and array without using methods and properties (using Exception Handling)

gaurav-4uk:
…es (using Exception Handling)

image

image


harshvegada (approved):
**`Approved`**
**`Code Looks good to me`**

PR #3002, opened 2021-10-01
ArrayList with normal for loop and enhanced for loop

gaurav-4uk:
image


harshvegada (approved):
**`Approved`**
**`Code Looks good to me`**

PR #3003, opened 2021-10-01
Precilla_Assignment44

precillaaranha16:
image

### DIFF
--- a/src/precilla/assignment44/FindingMaxNumber.java
+++ b/src/precilla/assignment44/FindingMaxNumber.java
@@ -1,0 +1,28 @@
+/*
+Assignment - 44 : 1st  Oct'2021 
+Find out the max number from an ArrayList.
+ */
+package precilla.assignment44;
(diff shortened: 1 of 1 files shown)


src/precilla/assignment44/FindingMaxNumber.java
```diff
@@ -0,0 +1,28 @@
...
+		System.out.println("Find out the max number from an ArrayList");
+		System.out.print("Input:[");
+		ArrayList<Integer> al=new ArrayList<Integer>();
+		al.add(10);
+		al.add(7);
+		al.add(99);
+		al.add(53);
+		al.add(74);
+		int max=al.get(0);
+		for(int index=0;index<al.size();index++) {
```
harshvegada, line 18: this should starts from 1 as you taken max variable which hold 0's index value.

harshvegada (approved):
**`Approved`**
**`Code Looks good to me`**

PR #3008, opened 2021-10-01
Purshottam_Assignment44

purshottam94:
image

### DIFF
--- a/src/purshottamJoshi/Assignment44.java
+++ b/src/purshottamJoshi/Assignment44.java
@@ -1,0 +1,24 @@
+package purshottamJoshi;
+
+import java.util.*;
+
+public class Assignment44 {
+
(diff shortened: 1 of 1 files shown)


src/purshottamJoshi/Assignment44.java
```diff
@@ -0,0 +1,24 @@
+package purshottamJoshi;
+
+import java.util.*;
+
+public class Assignment44 {
+
+	Object getMaxNumber(ArrayList arr) {
+		/*	Comparator c = null;
+		arr.sort(c);*/
+		Collections.sort(arr);
```
harshvegada, line 10: you should write raw logic instead of direct sorting.

harshvegada (approved):
**`Approved`**
**`Code Looks good to me`**

PR #3029, opened 2021-10-01
Ritabrata_Assignment_44

Ritabrata1996:
image

### DIFF
--- a/src/ritabrata2/Assignment44.java
+++ b/src/ritabrata2/Assignment44.java
@@ -1,0 +1,25 @@
+/*Assignment - 44 : 1st  Oct'2021 
+Find out the max number from an ArrayList.
+input: [10,7,99,53,74]
+output: 99*/
+
+package ritabrata2;
+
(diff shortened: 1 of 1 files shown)


src/ritabrata2/Assignment44.java
```diff
@@ -0,0 +1,25 @@
...
+public class Assignment44 {
+
+	public static void main(String[] args) {
+		ArrayList<Integer> al= new ArrayList<Integer>();
+		al.add(new Integer("10"));
+		al.add(new Integer("7"));
+		al.add(new Integer("99"));
+		al.add(new Integer("53"));
+		al.add(new Integer("74"));	
+		Object obj= Collections.max(al);
```
harshvegada, line 20: Instead of using predefined methods, you should use write logic.

harshvegada (approved):
The code looks good to me!
Yet, I think it's helpful to know how to approach an interview.

PR #3099, opened 2021-10-04
Shraddha_Assignment44

shraddhaatal:
image


nehatawari:
Logic is correct.u can create method and pass array list to that method instead of writing logic in main

PR #3244, opened 2021-10-12
Priti_Assingment_49

psnilange:
image
image


src/pritihascode_Equals/Builder.java
```diff
@@ -0,0 +1,73 @@
+package pritihascode_Equals;
+public class Builder{
```
harshvegada, line 2: you should implement Comparable interface

harshvegada (approved):
**`Approved`**

PR #3251, opened 2021-10-12
Indu_Assignment_49

InduBhartiya:
image

### DIFF
--- a/src/indu/Assignment_49/Builder.java
+++ b/src/indu/Assignment_49/Builder.java
@@ -1,0 +1,56 @@
+/*ssignment - 49 : 10th Oct [hashCode & Equals]
+
+Create a class call builder.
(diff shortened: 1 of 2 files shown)


src/indu/Assignment_49/Builder.java
```diff
@@ -0,0 +1,56 @@
...
+		this.name = name;
+		this.gstNumber = gstNumber;
+		this.totalCompleteProjects = totalCompleteProjects;
+		this.revenue= revenue;
+	}
+	
+	@Override
+	public boolean equals(Object obj)
+	{
+		Builder b1 = (Builder) obj;
```
harshvegada, line 34: You should implement Comparable then write @override method

harshvegada (approved):
**`Approved`**

PR #3292, opened 2021-10-14
Shraddha_assignment53_PateernPrintingAssignment

shraddhaatal:
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

image


harshvegada (approved):
you should create a separate method which prints this pattern, rest code looks good to me

